PR #1, opened 2017-12-29
Added proper headers to docs files and fixed all links

waquidvp:
Other things:
- Removed unused image files
- Most files were reformatted with Prettier
- Added Github Link to header
- Removed API tab from header
- Removed example docs file generated by Docusaurus


brentvatne:
hey @waquidvp - thanks a lot for this! unfortunately I think it will require a lot of work to rebase because I had some local changes that I hadn't pushed that impact about 10 of the files in the diff. this is why I suggested that you work on the landing page, so that we wouldn't run into any conflicts as I was making big changes to the rest of it. it's probably best to focus on the landing page at the moment as I'll continue to make big changes to the rest of the docs over the next few days

waquidvp:
No problem, the only reason I did it was cause it was an easy thing to start off with. Will focus my effort on the landing page from now on.

PR #6, opened 2018-01-18
StatusBar Config Docs

spencercarli:
Got most of the way but ran into an issue with the TabNavigator that I've never actually had to deal with. What if the last tab has a different statusbar config than the first? Since it's the last one rendered its statusbar config overrides the current screen's

tabbar-issue-demo


spencercarli:
I should be able to accomplish this pretty easily using the new events system.

spencercarli:
Apologies for the delay - just updated the docs to use events system for tab navigator.

docs/status-bar.md
```diff
@@ -2,4 +2,153 @@
 id: status-bar
 title: Different status bar configuration based on route
 sidebar_label: Different status bar configuration based on route
----
\ No newline at end of file
+---
+
+If you don't have a navigation bar, or your navigation bar changes color based on the route, you'll want to ensure that the correct color is used for the content.
```
brentvatne: maybe we should use the term "navigation header" for consistency with our api

docs/status-bar.md
```diff
@@ -2,4 +2,153 @@
...
+    screen: Screen2,
+  },
+});
+```
+
+![DrawerNavigator with different StatusBar configs](./assets/statusbar/statusbar-drawer-demo.gif)
+
+## TabNavigator
+
+If you're using a TabNavigator it's a bit more complex because the screens on all of your tabs are rendered at once - that means that the last `StatusBar` config you set will be used (likely on the final tab of your tab navigator, not what the user is seeing).
```
brentvatne: yeah this is very unfortunate. looking forward to integrating status bar style into react-nav itself

PR #14, opened 2018-02-03
Document new navigation actions and helpers

ericvicenti:


ericvicenti:
oh @react-navigation-bot, you so fancy!

PR #21, opened 2018-02-05
Fix broken links

n1ru4l:


brentvatne:
thanks! I found other broken links and fixed them all in a separate commit

PR #24, opened 2018-02-05
Fix order of parameter in getStateForAction

Axlle:
Simple typo.

StackRouter source code [here](https://github.com/react-navigation/react-navigation/blob/fae1f71fb3f3fcbf7ad8ffdb9e0f5f1a840bf660/src/routers/StackRouter.js#L101).


brentvatne:
maybe I'm missing something here -- the source code that you linked to shows `getStateForAction(action, state)`, the current docs show `getStateForAction(action, state)` and your proposed change would have the docs show `getStateForAction(state, action)`

Axlle:
My bad. I'm used to looking at the `(state, action)` pairs in Redux reducers so my brain got confused.

PR #51, opened 2018-02-13
Fix links to anchors

abhishek-raj:
The anchors created from the markdown are in lower case. I changed all links to anchors to lower case.

### DIFF
--- a/website/core/Footer.js
+++ b/website/core/Footer.js
@@ -28,7 +28,7 @@ class Footer extends React.Component {
                 this.props.config.baseUrl +
                 'docs/' +
                 // this.props.language +
-                '/getting-started.html'
+                'getting-started.html'
               }>
               Getting Started
             </a>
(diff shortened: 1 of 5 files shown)


abhishek-raj:
Seems like we aren't using the translation features of docusaurus and thus the links in the footer isn't working. Fixed those too.

website/core/Footer.js
```diff
@@ -27,35 +27,35 @@ class Footer extends React.Component {
               href={
                 this.props.config.baseUrl +
                 'docs/' +
-                this.props.language +
-                '/getting-started.html'
+                // this.props.language +
```
brentvatne: we should leave this in. it produces a slightly weird url right now but it still works, and we'll add translations shortly

abhishek-raj:
Reverted back the changes to Footer.js.

website/core/Footer.js
abhishek-raj: Okay. Reverted back the Footer.js

brentvatne:
thanks!

PR #54, opened 2018-02-13
Improve addListener documentation

slorber:
see also https://github.com/react-navigation/rfcs/issues/11

### DIFF
--- a/docs/navigation-prop.md
+++ b/docs/navigation-prop.md
@@ -108,6 +108,45 @@ React Navigation emits events to screen components that subscribe to them:
 - `didFocus` - the screen focused (if there was a transition, the transition completed)
 - `didBlur` - the screen unfocused (if there was a transition, the transition completed)
 
+
+Example:
+
+```javascript
+const didBlurSubscription = this.props.navigation.addListener('didBlur', payload => {
+  console.debug("didBlur",payload);
+});
+
+// Remove the listener when you are done
+didBlurSubscription.remove();
+```
+
+The payload will be a json object:
+
+```javascript
+{
+  action: {type: "Navigation/COMPLETE_TRANSITION", key: "StackRouterRoot"},
+  context: "id-1518521010538-2:Navigation/COMPLETE_TRANSITION_Root",
+  lastState: undefined,
+  state: undefined,
+  type: "didBlur",
+}
+```
+
+You can use this listener API to create async navigation actions so that you know when navigation has completed:
+
+```javascript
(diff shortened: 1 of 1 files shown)


docs/navigation-prop.md
```diff
@@ -108,6 +108,45 @@ React Navigation emits events to screen components that subscribe to them:
 - `didFocus` - the screen focused (if there was a transition, the transition completed)
 - `didBlur` - the screen unfocused (if there was a transition, the transition completed)
 
+
+Example:
+
+```javascript
+const didBlurSubscription = this.props.navigation.addListener('didBlur', payload => {
+  console.debug("didBlur",payload);
```
brentvatne: can you format this code with prettier? you can copy and paste it into snack to do that

docs/navigation-prop.md
```diff
@@ -108,6 +108,45 @@ React Navigation emits events to screen components that subscribe to them:
...
+});
+
+// Remove the listener when you are done
+didBlurSubscription.remove();
+```
+
+The payload will be a json object:
```
brentvatne: s/json/JSON

docs/navigation-prop.md
```diff
@@ -108,6 +108,45 @@ React Navigation emits events to screen components that subscribe to them:
...
+{
+  action: {type: "Navigation/COMPLETE_TRANSITION", key: "StackRouterRoot"},
+  context: "id-1518521010538-2:Navigation/COMPLETE_TRANSITION_Root",
+  lastState: undefined,
+  state: undefined,
+  type: "didBlur",
+}
+```
+
+You can use this listener API to create async navigation actions so that you know when navigation has completed:
```
brentvatne: I wonder if this part is better suited for some specific use case in the guide. The purpose of this section is an API reference

docs/navigation-prop.md
slorber: done

docs/navigation-prop.md
slorber: ???

docs/navigation-prop.md
slorber: removed

slorber:
not sure what you mean by `s/json/JSON`, tell me if it's good

brentvatne:
oh, `s/x/y` is a shorthand for saying replace x with y - it's how you do it in vim

slorber:
ah ok, I only know how to exit it ;)

PR #55, opened 2018-02-13
fix(getting-started): remove unused code

vvo:
This fixes https://twitter.com/EricVicenti/status/963541803537281024

Unsure you want to keep the content, it's in git still


ericvicenti:
Yay @vvo, thank you for getting to the bottom of this! If we want this content back we can revive it later

PR #68, opened 2018-02-23
Adding createNavigationReducer redux doc

yordis:
Related to https://github.com/react-navigation/react-navigation-redux-helpers/pull/5

@Ashoat I have another PR I want to make for remove more complexity upfront for basic users. That is why I create the section under the hood so advance users could read how to modify everything for their needs.

### DIFF
--- a/docs/redux-integration.md
+++ b/docs/redux-integration.md
@@ -45,20 +45,14 @@ import {
 import {
   createReduxBoundAddListener,
   createReactNavigationReduxMiddleware,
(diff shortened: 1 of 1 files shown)


yordis:
@Ashoat probably it is better to update the documentation all at once when the other PRs get in 😄

brentvatne:
@Ashoat @yordis - what's the status of this?

yordis:
@brentvatne the feature has been release already.

PR #74, opened 2018-03-01
Grammar

lustigdev:


brentvatne:
hm my grammar isn't perfect but the original reads well for me. thanks anyways!

Issue #32, opened 2018-02-07
Indicate in iPhone X section that people should use React Navigation SafeAreaView


brentvatne:
https://reactnavigation.org/docs/handling-iphonex.html
Related: https://github.com/react-navigation/react-navigation/issues/3173

PR #92, opened 2018-03-17
Document forceInset prop

jozan:
There has been some questions how to use `SafeAreaView` and previously it has not been made clear that importing it from `react-navigation` is preferred.


jozan:
This PR also closes this #32 .

brentvatne:
thanks @jozan!

PR #98, opened 2018-03-21
Normalizing and fixing links

richgilbank:
Currently the links are a bit inconsistent in formatting. In some cases, (eg. click `router` [here](https://reactnavigation.org/docs/custom-navigators.html#createnavigator)), this results in the links being broken. 

It looks like links pointing to API pages (eg. `StackNavigator` [here](https://reactnavigation.org/docs/custom-navigator-overview.html#built-in-navigators)) don't require `.html`, but links pointing to guide pages (like above) do. 

Either way, for now this normalizes the links in a format where they all work (`/docs/[page].html`). 

Thanks for all the work on the project! 😄


brentvatne:
thanks for this!

PR #101, opened 2018-03-22
Fixed navigation hierarchy

DarkSector:
All StackNavigator specific actions are discussed after the common navigation actions. Removed link to init as it is missing documentation (since it is confusing)


brentvatne:
heya, sorry for not getting back to you on this. we ended up splitting stack actions into their own file which I believe accomplishes the goal here

Issue #131, opened 2018-04-13
Custom navigators doc - addListener in child navigation


endoooo:
in order to support StackNavigator and DrawerNavigator as sub-navigators of a custom navigator, we need to supply `addListener` to the childNavigation prop.

today, the documented example considers only state and dispatch:

``` js
// The state of the active child screen can be found at routes[index]
let childNavigation = { dispatch, state: routes[index] };
// If we want, we can also tinker with the dispatch function here, to limit
// or augment our children's actions
```

I'll send a pull request to fix this and prevent others to spend hours trying to figure out why the sub-navigators just won't work... 😬

ericvicenti:
Thank you!

PR #132, opened 2018-04-13
Support for children Stack and Drawer navigators in custom navigator example

endoooo:
fix to #131

### DIFF
--- a/docs/custom-navigators.md
+++ b/docs/custom-navigators.md
@@ -10,16 +10,17 @@ A navigator is any React component that has a [router](https://github.com/react-
 class MyNavigator extends React.Component {
   static router = MyRouter;
   render() {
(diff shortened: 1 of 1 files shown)


endoooo:
uh... I've just noticed that with this "fix", the app will run but the navigation lifecycle listeners in the child navigators won't work as expected.

looking at the CardStack implementation, I've noticed that there is a significant amount of work to manage the child event subscribers. considering this, I have 2 questions:

1. there is a helper or anything else exported in React Navigation that can help with that (taking care of children navigators events subscriptions)?

2. can someone give an overview about how the navigation events and addListener works?

if there's no helper, I can see if I can do something in the main repo. I'll just need time to look closer at the addListener implementation 😃

PR #140, opened 2018-04-24
Reference createSwitchNavigator in the auth-flow.md

russelh15:
createSwitchNavigator was not being referenced in the setup guide.


brentvatne:
thanks!

PR #146, opened 2018-04-30
Migrate StackActions to a new page

richgilbank:
Since https://github.com/react-navigation/react-navigation/pull/3619, actions for stack navigators have moved to the `StackActions` object, so the documentation for `replace` and `reset` needed an update. 
For the most part, it's a direct swap of the StackActions content into a new page, with a few additional changes:
* Examples now show `StackActions` being imported and used as well as `NavigationActions`
* Fixed italics on prop types in a few examples
* Added an intro for StackActions and links between the two pages. 

Happy to make additional changes if there's anything missing/unclear. Thanks again for all your work on this project! 

_Note_: I haven't yet added additional docs for `Push`, `Pop` or `PopToTop`.

### DIFF
--- a/docs/navigation-actions.md
+++ b/docs/navigation-actions.md
@@ -15,13 +15,7 @@ The following actions are supported:
 * [Set Params](#setparams) - Set Params for given route
 * [Init](#init) - Used to initialize first state if state is undefined
 
-Within a stack, you can also use:
-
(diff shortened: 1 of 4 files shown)


brentvatne:
thanks! this looks good to me. if you're willing to submit a follow up to add `Push`, `Pop`, `PopToTop` that'd be great :)

PR #147, opened 2018-04-30
Descriptions for StackActions.{push|pop|popToTop} (with examples!)

richgilbank:
Followup to https://github.com/react-navigation/react-navigation.github.io/pull/146


richgilbank:
FWIW, here's what it looks like rendered: 

image

PR #149, opened 2018-05-01
Update switch-navigator to explain the use case

AuroreM:
**Motivation** : I was reading the documentation, and I could not understand the difference between `SwitchNavigator` and `StackNavigator`. I has to look for on StackOverflow and I find [this answer](https://stackoverflow.com/questions/49417757/what-is-the-difference-between-react-navigations-stacknavigator-and-switchnavig). I though it would be a could idea to have it in the documentation.

**Look**:
image

### DIFF
--- a/docs/switch-navigator.md
+++ b/docs/switch-navigator.md
@@ -4,8 +4,13 @@ title: createSwitchNavigator
 sidebar_label: createSwitchNavigator
 ---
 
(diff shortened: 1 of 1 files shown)


AuroreM:
@brentvatne @vonovak Have you got a chance to look at it ? :)

vonovak:
No worries, the PR is not forgotten. We will get to it :)

brentvatne:
looks great, thanks!

AuroreM:
Thanks @vonovak @brentvatne, glad to contribute :)

PR #153, opened 2018-05-08
link broken..

nonotest:


brentvatne:
thanks, fixed in https://github.com/react-navigation/react-navigation.github.io/commit/c7691b279358200b72987509b0d838b2074c8766

PR #154, opened 2018-05-08
Dispatch popAction in `stack-actions.md`

lizraeli:
In `stack-actions.md` line 99, popAction rather than pushAction should be dispatched


brentvatne:
thanks!

PR #186, opened 2018-06-11
Enhance the doc about tabBarComponent

Albert-Gao:
Related to the discussion here: https://github.com/react-navigation/react-navigation/issues/4297


brentvatne:
I'd rather see a heading in this doc that explains the use case and the solution, I think having this comment inline doesn't provide enough context

Albert-Gao:
Yes, that would be clearer, do you mean, add a heading in the top area? Just for this issue? Or add an example for the `tabBarComponent`?

brentvatne:
another heading in this same docs page would be good

Albert-Gao:
Add an example below :)

PR #193, opened 2018-06-13
Update Redux Integration docs for react-navigation-redux-helpers@2.0.0

Ashoat:
Introducing `reduxifyNavigator`, which should make things much simpler.


brentvatne:
deployed, thanks!

PR #243, opened 2018-09-11
Update material-top-tab-navigator.md

ralph-dev:
**Motivation**
There is no easy way to center the indicator or any supporting documentation. I suggest a re-write of the component to not use Absolute for positioning but this will do for now in helping the majority of people (As seen in the official Discord channel who could still not solve this).

### DIFF
--- a/docs/material-top-tab-navigator.md
+++ b/docs/material-top-tab-navigator.md
@@ -54,7 +54,7 @@ tabBarOptions: {
     width: 100,
   },
   indicatorStyle: {
-    width: 100
+    width: 100,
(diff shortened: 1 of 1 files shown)


brentvatne:
hello! I think this isn't very clear for people who are reading the docs without any other context. you could put a separate subsection on the page, maybe something like Workarounds and then a heading for this particular one. i don't really understand what centering the indicator means here either, so you may want to elaborate further on that. if this is a common feature then i'd go and file a feature request on react-native-tab-view or open a PR there to implement it properly

Issue #236, opened 2018-08-27
Guide to linking with hybrid iOS apps


James-E-Adams:
Hi there!

Thanks for a great library :)

I found it a bit hard to stumble on how to get `react-navigation` working in a hybrid app. 

It just requires you adding the `RCTLinkingIOS` subspec in your `podfile` under React and running `pod install`, a la: https://github.com/react-navigation/react-navigation/issues/2977

Is there somewhere can document this? Happy to open a PR if someone could inform me where this sort of documentation should live!

Thanks

brentvatne:
hi @James-E-Adams - wow I don't know how I missed this for so long. It does indeed require `RCTLinkingIOS` to be added - this is the default behavior when initializing a new project but I imagine in hybrid apps you may not opt-in to including it. A pull request for this is welcome! I also added a "Hacktoberfest" tag to this issue :P ([see more](https://hacktoberfest.digitalocean.com))

James-E-Adams:
Nice! I just signed up for Hacktoberfest so I'll look at this tomorrow potentially? :)

James-E-Adams:
Done. https://github.com/react-navigation/react-navigation.github.io/pull/259

PR #259, opened 2018-10-01
docs(getting-started): Add explanation for hybrid iOS apps

James-E-Adams:
Adds docs as per #236


brentvatne:
thank you! looks good

Issue #197, opened 2018-06-14
Document safeAreaInset for createBottomTabNavigator 


brentvatne:
https://github.com/react-navigation/react-navigation-tabs/pull/32

PR #260, opened 2018-10-03
Documents the `safeAreaInset` prop for `<TabBarBottom>`

bradbumbalough:
- Closes https://github.com/react-navigation/react-navigation.github.io/issues/197


brentvatne:
thank you!

PR #270, opened 2018-10-19
State Persistence Blog Link Updated

mdchristians:
Updated the blog link to the correct state-persistence doc URL.


brentvatne:
thanks!

PR #276, opened 2018-10-23
Add reminder to include super(props)

ryanfs:
The example doesn't include this dependency and people will have to remember to add it on their own. It would be useful to have a fully working Button class in the example. I pasted the example, forgot to add the constructor function on my first go around and only remembered to add it in after getting the error again.


ryanfs:
This PR is a suggestion to improve the docs.

ryanfs:
@vonovak my mistake you are right, adding `super(props)` is not necessary. I had a different issue earlier that lead me to believe it was.  Thanks for the response! Closing this PR now.

PR #277, opened 2018-10-24
Cross linking back to guides

DamonLC21:
While looking up a way to style the createBottomTabNavigator while being exported in React-Native I came across the API Reference page first.  The examples we're a bit confusing before discovering the guides as well.  I think it would be great if some cross linking could be implemented to further help those who may stumble across the API references first before seeing the guides.  Would save a lot of time and further improve some already awesome documentation for React-Navigation.


brentvatne:
thanks!

PR #307, opened 2018-11-26
Component name leads to expo throwing error

Staaikies:
Rootstack isn't mentioned in any of the other code. Causing expo to spit out an error.


brentvatne:
thanks, I updated the rest of the file for consistency

PR #326, opened 2018-12-20
Update deep-linking.md

anhtuank7c:
Update guide about handle "Dynamic Links" with react-navigation "deeplink" v3


brentvatne:
@anhtuank7c - thanks for this! could you possibly move the code to a separate repository or something so we can just link to it from here?

satya164:
Closing due to no response.

PR #328, opened 2018-12-21
Improve documentation of `lazy` property on TabNavigatorConfig

rgilbert1:
I added my changes to `bottom-tab-navigator` in the version 3.x docs, I hope that was the right thing to do 😬


website/versioned_docs/version-3.x/tab-navigator.md
```diff
@@ -0,0 +1,126 @@
+---
+id: version-3.x-tab-navigator
+title: createTabNavigator
+sidebar_label: createTabNavigator
```
brentvatne: createTabNavigator is gone in react-navigation@>=3.0, we have `createBottomTabNavigator` and `createMaterialTopNavigator` instead now :) so we can remove this doc

brentvatne:
aside from that one comment i left it looks good to me! let me know when that change is ready and i'll land it

website/versioned_docs/version-3.x/tab-navigator.md
rgilbert1: Got it, makes sense that the file wasn't there already then!

rgilbert1:
Thanks Brent! We should be good to go!

brentvatne:
thank you @rgilbert1! 😄

Issue #191, opened 2018-06-12
Guide around doing some action after re-focusing a screen


brentvatne:
Cover both `withNavigationFocus` and `addListener`

khushboogupta1:
@brentvatne Thanks for answering.. but as  i am new to react native not getting where to use this. Please help me with some example.

khushboogupta1:
@MrPluto  i am using stack navigator. According to your answer what i did is  i called my API in 'didFocus', () { "HERE API CALL " }. But still no luck..

MrPluto:
show some code ?

khushboogupta1:
@MrPluto  Mailed you

MarcelZOI:
Hey Girls and Guys,

i have also an problem with redux and react-navigation:

here the structure:

<App>
  <Main>
    <Header/> //this header has an number that displays the items inside the shopping cart value={this.props.cart.elements.length}
    <MainStack>
       <Home>
       <ArtList>
       <ShoppingCart>
    </MainStack>
  </Main>
  <AddCart/> //Modal when saving the Article to the cart it goes Back to the MainStack (navigation.goBack())
</App>

So if i add an item to the cart, the counter inside the header didnt changed. After switching between the stack-screens the number get updated.

i have the latest version of redux, react-navigation and react-native.

Any ideas?

MarcelZOI:
@khushboogupta1  i solved it, maybe you also can use it:

i have a NavigationService where i handle all my NavigationActions (e.g. goBack(), navigate()...)
(https://reactnavigation.org/docs/en/navigating-without-navigation-prop.html)

when set the _navigator i also pass a function (in this case _goBackEvent).

```js
function setTopLevelNavigator(navigatorRef, goBackEvent) {
  _navigator = navigatorRef;
  _goBackEvent = goBackEvent;
}
```

and this function i called in the goback function:

```js
function goBack(){
  _navigator.dispatch(
    NavigationActions.back()
  )
  if (_goBackEvent)
    _goBackEvent();
}
```

in my class where i added the Stack:

```js
 <StackNav           
    ref={navigatorRef => {
      NavigationService.setTopLevelNavigator(navigatorRef, () => {
        //do anything here when goBack is triggered
      });
    }}
  />
```

RodolfoSilva:
In my case i use `willFocus`:
```javascript
componentDidMount() {
  this.subs = [
    this.props.navigation.addListener('willFocus', () => {
    }),
  ];
}

[...]

PR #330, opened 2018-12-28
Adds guide: action on focus

jonsamp:
Adds a guide: **Call an Action After Focusing Screen**

This is in response to this Help Wanted issue: https://github.com/react-navigation/react-navigation.github.io/issues/191

See the guide here: https://deploy-preview-330--react-navigation.netlify.com/docs/en/action-after-focusing-screen.html

### DIFF
--- a/docs/action-after-focusing-screen.md
+++ b/docs/action-after-focusing-screen.md
@@ -1,0 +1,85 @@
+---
+id: version-3.x-action-after-focusing-screen
+title: Call an Action After Focusing Screen
+sidebar_label: Action After Focusing Screen
+original_id: action-after-focusing-screen
+---
+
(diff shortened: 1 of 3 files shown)


brentvatne:
thanks @jonsamp, this looks great! a few notes:

- before merging we should create an identical file in the unversioned docs directory
- it may be worth mentioning that using the HOC will re-render the screen when isFocused changes unless you implement shouldComponentUpdate. so if you want to make an api call it may be better to use the listener instead, since that won't re-render until you do a setState after the api call completes.
- it may also be worth mentioning that another use case may be to pause some resource (eg: watching location, playing a video) when switching away from the screen

jonsamp:
thanks @brentvatne for the feedback! i incorporated your feedback and also copied the markdown file to the **docs/** directory. please let me know if you have more feedback!

brentvatne:
looks great, thanks @jonsamp :)

Issue #81, opened 2018-03-09
Add badge to tab bar


brentvatne:
https://github.com/react-navigation/react-navigation/issues/3692

Seems like a reasonable thing to document

PR #331, opened 2018-12-28
Add tutorial of adding badges to tabbar icons

sunnylqm:
Resolves https://github.com/react-navigation/react-navigation.github.io/issues/81
Snack preview: https://snack.expo.io/@react-navigation/tabs-with-icons-v3

### DIFF
--- a/docs/tab-based-navigation.md
+++ b/docs/tab-based-navigation.md
@@ -64,16 +64,19 @@ export default createBottomTabNavigator(
     defaultNavigationOptions: ({ navigation }) => ({
       tabBarIcon: ({ focused, horizontal, tintColor }) => {
         const { routeName } = navigation.state;
+        let IconComponent = Ionicons;
         let iconName;
         if (routeName === 'Home') {
           iconName = `ios-information-circle${focused ? '' : '-outline'}`;
+          // Sometimes we want to add badges to some icons. 
+          // You can check the implementation below.
+          IconComponent = HomeIconWithBadge; 
         } else if (routeName === 'Settings') {
           iconName = `ios-options${focused ? '' : '-outline'}`;
         }
 
-        // You can return any component that you like here! We usually use an
-        // icon component from react-native-vector-icons
-        return <Ionicons name={iconName} size={horizontal ? 20 : 25} color={tintColor} />;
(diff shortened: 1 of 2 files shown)


website/versioned_docs/version-3.x/tab-based-navigation.md
```diff
@@ -85,14 +88,61 @@ export default createBottomTabNavigator(
...
+```js
+export default class IconWithBadge extends React.Component {
+  render() {
+    const { name, badgeCount, color, size } = this.props;
+    return (
+      <View style={{ width: 24, height: 24, margin: 5 }}>
+        <Ionicons name={name} size={size} color={color} />
+        { badgeCount > 0 && (
+          <View style={{
+            // Absolute positioning ref: https://developer.mozilla.org/en-US/docs/Web/CSS/position#Absolute_positioning
```
brentvatne: can we condense this comment to one line? maybe just `// If you're using react-native < 0.57 overflow outside of parent will not work on Android, see https://git.io/fhLJ8`

docs/tab-based-navigation.md
```diff
@@ -84,14 +87,61 @@ export default createBottomTabNavigator(
 );
 ```
 
-<a href="https://snack.expo.io/@react-navigation/tabs-with-icons-v3" target="blank" class="run-code-button">&rarr; Run this code</a>
+<a href="https://snack.expo.io/@sunnylqm/tabs-with-icons-v3" target="blank" class="run-code-button">&rarr; Run this code</a>
```
brentvatne: I updated the code here to clean it up a little bit: https://snack.expo.io/@react-navigation/tabs-with-icons-v3

can you update the copy of it in these docs to match?

website/versioned_docs/version-3.x/tab-based-navigation.md
brentvatne: I updated the comment in https://snack.expo.io/@react-navigation/tabs-with-icons-v3 let me know if that works for you

sunnylqm: Sure. Updated.

brentvatne:
thanks, this should be helpful for people :)

PR #333, opened 2018-12-30
Document modal that dismisses by gesture only when ScrollView is at top

### LINKED ISSUE #315: Document how to build a modal that dismisses by gesture only when ScrollView is at top
https://github.com/react-navigation/react-navigation/issues/5320

ryan-gilb:
This fixes #315!

Any and all feedback is welcome :)

### DIFF
--- a/docs/custom-modal-dismissal.md
+++ b/docs/custom-modal-dismissal.md
@@ -1,0 +1,89 @@
+---
+id: custom-modal-dismissal
+title: Customizing modal dismissal
+sidebar_label: Customizing modal dismissal
+---
+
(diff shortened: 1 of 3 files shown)


rgilbert1:
Oh man, I initially didn't notice the delay, but now that you point it out it's awful! You're right, we can do better. I'll take a look at that TransparentStack!

Thanks for the feedback :)

jfrolich:
Hmm any pointers how to build this interaction? This proposed solution doesn't really work every time. This would be nice to document, as a scrollable + modal is a pretty common use case, and it seems to be possible now that RN3.0 uses `react-native-gesture-handler`. I don't really know where to start however.

zeh:
@jfrolich Ditto. @ryan-gilb did you manage to ever fix the issue where pressing down sometimes works (starting the modal close gestures), sometimes doesn't (still pulling from the scrollview)?

PR #334, opened 2018-12-31
 Adjust deep link set up and usage on Android

adrianha:
This PR adjust deep link set up on Android so the prefix will be the same with iOS.
I also add `android:launchMode` on set up instruction in order to support deep link navigation when the app is running.

### DIFF
--- a/docs/deep-linking.md
+++ b/docs/deep-linking.md
@@ -93,9 +93,7 @@ Next, let's configure our navigation container to extract the path from the app'
 ```js
 const SimpleApp = createAppContainer(createStackNavigator({...}));
 
-// on Android, the URI prefix typically contains a host in addition to scheme
-// on Android, note the required / (slash) at the end of the host property
-const prefix = Platform.OS == 'android' ? 'mychat://mychat/' : 'mychat://';
+const prefix = 'mychat://';
 
 const MainApp = () => <SimpleApp uriPrefix={prefix} />;
 ```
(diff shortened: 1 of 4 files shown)


docs/deep-linking.md
```diff
@@ -141,19 +139,26 @@ To test the URI on a real device, open Safari and type `mychat://chat/Eric`.
...
-<intent-filter>
-    <action android:name="android.intent.action.VIEW" />
-    <category android:name="android.intent.category.DEFAULT" />
-    <category android:name="android.intent.category.BROWSABLE" />
-    <data android:scheme="mychat" android:host="mychat" />            
-</intent-filter>
+<activity
+    android:name=".MainActivity"
+    android:launchMode="singleTask"
+    ...>
```
brentvatne: ideally code should be copy+pasteable from docs, so we should remove the `...` here

ericvicenti:
I'm not super qualified to comment on Android manifest configuration.. but the last time I tested with a bare RN project, the additional manifest configuration was not needed for deep linking to work.

Are these additional steps mandatory now, or do they improve upon the base behavior? In any case it would be nice to have a bit more explanation in the docs to explain what is happening and why.

On the topic of "nice-to-haves", it would be awesome if @adrianha or anybody else wanted to make a bare example ReactNav app that demonstrates deep linking configuration. We could link to it from the docs, and perhaps put it under the ReactNav org.

docs/deep-linking.md
adrianha: Updated

adrianha:
hi @brentvatne @ericvicenti,

it is true that we don't need to add manifest configuration to make the deep link works.
Additional manifest configuration is needed when you want to perform in app navigation using your deep link. This configuration will make user be able to navigate to `ChatScreen` by using `Linking` API like: `Linking.openURL('mychat://chat/Eric');`

adrianha:
Do i get the ticket for App.js conf @brentvatne ? 😂

brentvatne:
yes @adrianha! dm me on twitter :)

PR #365, opened 2019-02-17
Updated v3 docs for drawerOpenProgress

rgoldiez:
Updated v3 docs to include drawerOpenProgress information and example usage.

cc @brentvatne


brentvatne:
looks great, thanks!

Issue #378, opened 2019-03-13
Tracking page is outdated


gzzo:
In https://reactnavigation.org/docs/en/screen-tracking.html the onNavigationStateChange is a prop of an appContainer, not of a stackNavigator

PR #379, opened 2019-03-13
Update screen tracking page with appContainer

gzzo:
See #378 

While I'm editing this page, I believe all the redux content on this page should be removed.


brentvatne:
thanks!

> I believe all the redux content on this page should be removed.

I agree, could you update the PR accordingly?

brentvatne:
thanks!

PR #383, opened 2019-03-21
Remove nonexistent icon from tab-based-navigation.md

mstiggle:
`tab-based-navigation.md` was trying to use `ios-options-outline` from `Ionicons` which doesn't exist. Removed that code so that now the `Settings` icon in the example only uses `ios-options`.


brentvatne:
thanks, that changed in a recent vector-icons update :)

PR #390, opened 2019-03-29
Add documentation for deep linking in nested navigators

### LINKED ISSUE #387: Explain how to use deep linking in nested navigators
This issue [https://github.com/react-navigation/react-navigation/issues/1527#issuecomment-439201168](https://github.com/react-navigation/react-navigation/issues/1527#issuecomment-439201168) explains that you need to add an empty path string to all parent navigators, or else the deep link opens the app, but won't link to the correct screen.

Unfortunately this isn't mentioned in the docs :)

Can someone help me how this should be added to the [deep linking docs](https://reactnavigation.org/docs/en/deep-linking.html)?  I thought either at the end of the _Configuration_ paragraph or as a new menu item.

benseitz:
This PR Fixes #387 by adding documentation to deep linking in nested navigators.

### DIFF
--- a/docs/deep-linking.md
+++ b/docs/deep-linking.md
@@ -29,6 +29,35 @@ const SimpleApp = createAppContainer(createStackNavigator({
 }));
 ```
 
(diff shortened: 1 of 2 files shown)


brentvatne:
thanks! this will save people some time for sure

PR #392, opened 2019-04-03
update tab nav example

quinlanj:
description below suggests that this should be `navigationOptions` instead of `defaultNavigationOptions`


brentvatne:
👌

Issue #373, opened 2019-03-07
Document enableURLHandling


brentvatne:
Useful to know prop on app container

jarvisluong:
Could you tell me what is this feature and how can it bring any benefits? 🤔

brentvatne:
as per https://reactnavigation.org/docs/en/deep-linking.html react-navigation handles deep links into your app by default. this isn't always desirable, and in v4 we are going to make it *not* do this by default. to disable automatic deep linking handling, you can pass `enableURLHandling={false}` to the app container

```jsx
const MyStack = createStackNavigator({ /* whatever */});
const Navigation = createAppContainer(MyStack);

export default class App extends React.Component {
  render() {
    return <Navigation enableURLHandling={false} />
  }
}
```

PR #418, opened 2019-04-22
add section to disable deep-linking

jarvisluong:
https://github.com/react-navigation/react-navigation.github.io/issues/373

I'm not sure if adding this info to that section is ideal... 🤔


brentvatne:
i think this is a reasonable place to put it! thanks

PR #420, opened 2019-04-23
Update tab navigator docs to use react-navigation-tabs

satya164:
Maybe we need to create a 4.x alpha and add these docs there? Though if tab and drawer versions are managed independently, not sure how we handle that.


brentvatne:
the docs directory should be a good place for this for now, then we can create a 4.x release from it when we are ready to ship it

satya164:
@brentvatne should we merge the PR then? or you want any changes?

brentvatne:
pr looks good as-is

PR #444, opened 2019-06-16
Update header-buttons.md

Yasser-G:
Setting navigation params in constructor instead of componentDidMount.


osdnk:
Hey,
Can you explain why you're doing this?

Yasser-G:
@osdnk  I really don't remember this PR, I think I made this to prevent re rendering the component, so it is better to use constructor instead of didMount.

satya164:
Side effects or updates shouldn't be done in constructor. Closing.

PR #453, opened 2019-07-08
Add docs for linking gesture handler on RN >= 0.60

satya164:
In the docs, I have added config to disable autolinking on both Android and iOS so that linking is easier with `react-native-link`. Otherwise the user needs to run `git checkout ios` to revert the changes which is another extra step there.

I haven't tested this. Just sent PR because people are having trouble and this is what I found.

cc @thymikee

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -32,7 +32,22 @@ yarn add react-native-gesture-handler
 # npm install react-native-gesture-handler
 ```
 
-Link all native dependencies:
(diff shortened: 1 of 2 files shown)


thymikee:
No need to disable ios. You can pass `--platforms android` param to link only Android.

satya164:
Maybe still worth disabling it? Otherwise we need to list 2 commands, one for React Native 0.60+ and one for below. If autolink worked for both platforms for gesture handler, these separate commands would've been avoidable.

PR #455, opened 2019-07-11
Fix getting started instructions

satya164:


docs/getting-started.md
```diff
@@ -84,6 +69,45 @@ public class MainActivity extends ReactActivity {
...
+      platforms: {
+        android: null,
+        ios: null,
+      },
+    },
+  },
+};
+```
+
+Use [jetifier](https://github.com/mikehardy/jetifier) to work around libraries using the support library on Android:
```
thymikee: How about rephrasing:

You can include it in your "postinstall" step of "package.json" file:
```json
{
  "postinstall": "npx jetifier"
}
```
This way you can avoid installation steps

PR #462, opened 2019-07-21
update formatting of Stack Actions

koenpunt:
`will` was incorrectly included in a code string, so I fixed that, and in the process I've also updated the formatting of several other code blocks.


brentvatne:
looks nicer :) thank you

PR #464, opened 2019-07-24
Provide better documentation with goBack(key)

### LINKED ISSUE #94: goBack reference use of keys is confusing
Hi, I noticed a discrepancy between the documentation and implementation of keys used for navigation, for example in the case of navigating backwards from a specific screen in the stack using goBack(key). The documentation, especially the code examples, make it seem like the key is the route name, however this is not the case and does not work. The documentation should be updated to reflect this, currently it is very misleading and the code examples provided do not work in practice.

bradbumbalough:
- Closes #94


brentvatne:
thanks!

PR #465, opened 2019-07-25
@tchayen/stack alpha additions

tchayen:
Adding a bit more details to #454.

Not sure about migration section, probably belongs more to the release docs, but on the other hand I believe that those changes will cause a lot of problems for some people and having it mentioned in the docs would be cool in my opinion.

Also, I am aware that prettier messed up with the samples. I think it is desirable, but let me know if you want to have it removed.


osdnk:
Please include onTransitionStart and onTransitionEnd params

tchayen:
@osdnk Sure. I spoke with @satya164 and he prefers having a parameter object instead of single argument there. Will include those changes as soon as the API gets closer to final.

PR #469, opened 2019-08-14
Add warning about remote debugging

allthetime:
Deep-linking does not function as expected when the remote debugger is enabled. If the app is already running on the phone (foreground or background) it works as expected, but if the app is closed opening a link will only ever open the initial route. This is an annoying and opaque issue to diagnose and it would be a boon to developers to be warned about it in the documentation.

### DIFF
--- a/docs/deep-linking.md
+++ b/docs/deep-linking.md
@@ -164,6 +164,8 @@ xcrun simctl openurl booted mychat://chat/Eric
 
 To test the URI on a real device, open Safari and type `mychat://chat/Eric`.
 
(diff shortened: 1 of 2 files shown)


satya164:
A lot of things actually don't work as intended when Remote debugging, not only deep linking. Maybe we should add this in a troubleshooting section

jarvisluong:
Sorry for a late reply. The information is already here in the documentation: https://reactnavigation.org/docs/troubleshooting/#app-is-not-working-properly-when-connected-to-chrome-debugger

PR #472, opened 2019-08-20
s/Device.hasNotch()/DeviceInfo.hasNotch()/

### LINKED ISSUE #471: Use collect name for `react-native-device-info`
I was reading the document of [Supporting safe areas](https://reactnavigation.org/docs/en/handling-iphonex.html) part to fix android specific notch, but found that sample code in [here](https://reactnavigation.org/docs/en/handling-iphonex.html#android-notches) for `hasNotch()` API is probably not collect?

```
import { Platform } from 'react-native';
import { SafeAreaView } from 'react-navigation';
import DeviceInfo from 'react-native-device-info';

if (Platform.OS === 'android' && Device.hasNotch()) { //DeviceInfo.hasNotch()/ is collect?
  SafeAreaView.setStatusBarHeight(
    /* Some value for status bar height + notch height */
  );
}
```

source is [here](https://github.com/react-native-community/react-native-device-info#hasNotch)

chansuke:
Fixes #471 

`DeviceInfo.hasNotch()` is collect?

source: [hasNotch()](https://github.com/react-native-community/react-native-device-info#hasNotch)

### DIFF
--- a/website/versioned_docs/version-3.x/handling-iphonex.md
+++ b/website/versioned_docs/version-3.x/handling-iphonex.md
@@ -42,7 +42,7 @@ export default createStackNavigator({
 To fix this issue you can wrap your content in a `SafeAreaView`, which can be imported from `react-navigation`. Recall that `SafeAreaView` should not wrap entire navigators, just the screen components or any content in them.
 
 ```jsx
-import { SafeAreaView } from 'react-navigation';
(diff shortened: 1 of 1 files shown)


brentvatne:
we don't recommend using this library for safe-areas anymore, will update docs soon. you should use this instead: https://github.com/th3rdwave/react-native-safe-area-context

PR #482, opened 2019-09-03
fix docs typo: screenOptopns => screenOptions

FLGMwt:
This doesn't seem to be an issue in other doc versions since they don't use the component-based API.


satya164:
Thank you for the PR

PR #483, opened 2019-09-03
docs typo: use defaultNavigationOptions tab navigator

FLGMwt:
This only seems to be an issue in v4 of the docs. v3 correctly uses `defaultNavigationOptions` and vNext uses a different API and had a separate issue (#482). Also worth noting that the linked snack correctly has `defaultNavigationOptions`. 

Totally didn't spend half an hour to figure out this was a bug in the docs 😬


brentvatne:
thanks!

PR #489, opened 2019-09-06
Added installation of `react-native-screens`

burhanyilmaz:
if not added this two lines code, the app crashed on android.
Related to this issue https://github.com/react-navigation/react-navigation/issues/6267.

### DIFF
--- a/docs/stack-navigator.md
+++ b/docs/stack-navigator.md
@@ -53,13 +53,6 @@ Next, we need to link these libraries. The steps depends on your React Native ve
   react-native link react-native-gesture-handler
   ```
 
-Installation of `react-native-screens` for Android add the following two lines to `dependencies` section in `android/app/build.gradle`:
-
-```
-implementation 'androidx.appcompat:appcompat:1.1.0-rc01'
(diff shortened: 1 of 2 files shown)


brentvatne:
it seems that this has become required as of react-native-screens@2.0.0-alpha :O can you add these instructions to https://reactnavigation.org/docs/en/getting-started.html#install-into-an-existing-project instead?

burhanyilmaz:
@brentvatne I added to getting started.

ahartzog:
@burhanyilmaz the flow of the getting started is still very unclear on whether these steps are required for 0.60+

Could I recommend adding a line...here (see image) that just says "FOR ALL VERSIONS"? Right now it appears it could be a subset of the 0.59 and lower instructions. 
image

burhanyilmaz:
@ahartzog According to SDK versions should work without problems. But it should be tested in rn versions.
You can check
**androidx.appcompat:appcompat:1.1.0-rc01**
https://androidx.tech/artifacts/appcompat/appcompat/1.1.0-rc01/
**androidx.swiperefreshlayout:swiperefreshlayout:1.1.0-alpha02**
https://androidx.tech/artifacts/swiperefreshlayout/swiperefreshlayout/1.1.0-alpha02/

PR #490, opened 2019-09-08
tabBarComponent usage error fixed.

abdurrahmanekr:
If I try to use it like in the document, I get the error "Cannot read property 'map' of undefined".


satya164:
Thanks for the PR

PR #491, opened 2019-09-14
Function call need to be make to create Ref

alexakasanjeev:
In code it is written
```js
export const navigationRef = React.createRef
```

But to create ref, functional call need to be made
```js
export const navigationRef = React.createRef();
```


alexakasanjeev:
Hi, should I create different pull requests containing 1 commit or 1 pull request containing multiple commits, because I have two opened pull request, each for different page

alexakasanjeev:
Thank you for accepting my pull request

PR #494, opened 2019-09-16
Add blog post for 4.x release

satya164:


website/blog/2019-09-16-react-navigation-4.0.md
```diff
@@ -0,0 +1,23 @@
+---
+title: React Navigation 4.0
+author: Core Team
+authorURL: https://twitter.com/reactnavigation
+---
+
+The documentation is now live at https://reactnavigation.org, and v3 lives [here](/docs/en/3.x/getting-started.html).
+
+In this release, we have removed the navigators from the `react-navigation` package. The navigators have lived in separate packages for a quite a while and you could already use those packages manually, but we still bundled them in the `react-navigation` package. This resulted in the bundled navigators not being up-to-date.
```
brentvatne: In this release, we have removed the navigators from the `react-navigation` package. The navigators have lived in separate packages for quite a while and you could already use those packages manually, but we still bundled them in the `react-navigation` package. This made it difficult for us to release significant updates to navigators, because we had to then do a major version release of `react-navigation` too. By separating the navigator packages there is more freedom to update and improve navigators without any impact on folks that don't use them.

For example, you will find when you install the latest versions of the drawer and tab navigators that the animations are more performant because they use `react-native-reanimated` to smoothly animate gestures. These have been available in `react-navigation-drawer` and `react-navigation-tabs` for several months now but we delayed updating them in `react-navigation` itself because we did not want to force every user to update.

PR #497, opened 2019-09-17
Add documentation for labelPosition option in createBottomTabNavigator

jsfroth:
Hi,

I noticed that the labelPosition options is missing in the documentation, so I added it :) I am not sure about to which versions I should add it in the documentation. It was released with @react-navigation/tabs 2.3.0, but the versions here in the website repo look like they refer to the versions of the main package. If you give me a hint which versions should be updated, I can update the PR.

Regards
Josef


satya164:
Thanks for the PR! You need to add them for 4.0

jsfroth:
You're welcome! Thanks for all your good work with react-navigation. I added another commit

PR #499, opened 2019-09-19
Fix typo in useIsFocused API docs

Un3qual:
The `title` and `sidebar_label` of the `useIsFocused` documentation were set to `useFocusEffect`. Hooks are only in the `next` version, so I don't think it needs to be updated anywhere else.


Un3qual:
Whoops. Just saw someone else already took care of this. Sorry

PR #514, opened 2019-09-21
remove ios13 beta status from themes pages (4.x, 3.x)

ivnnv:
This is all, just remove the text `(currently in beta)` from iOS 13 references on the themes pages


satya164:
Thanks for the PR!

PR #518, opened 2019-09-30
Fix(doc-4.x): Add required configuration

Rizwanjamal:
# Minor Addition In Doc: 4.x

After following the doc for `React Native 0.59 and lower`, mostly users getting error `Execution failed for task ':react-native-gesture-handler:compileDebugJavaWithJavac'` on running `react-native run-android`. I recently faced this issue as well. In order to resolve this issue we need to add required config values in `gradle.properties`. 
So in order to facilitate newbies & other users of this module. I've created this PR.
Please merge this PR, I will be really happy to be one of the contributor :+1:

### DIFF
--- a/website/versioned_docs/version-4.x/getting-started.md
+++ b/website/versioned_docs/version-4.x/getting-started.md
@@ -92,6 +92,13 @@ Next, we need to link these libraries. The steps depends on your React Native ve
   # npm install --save-dev jetifier
   ```
 
+  Next, add these two lines in `gradle.properties` under `android` folder
+
+  ```sh
+  android.useAndroidX=true
+  android.enableJetifier=true
+  ```
+
   Then add it to the `postinstall` script in `package.json`:
 
   ```json


website/versioned_docs/version-4.x/getting-started.md
```diff
@@ -92,6 +92,13 @@ Next, we need to link these libraries. The steps depends on your React Native ve
   # npm install --save-dev jetifier
   ```
 
+  Next, add these two lines in `gradle.properties` under `android` folder
```
satya164: cc @mikehardy can you verify if these instructions are correct?

mikehardy: @satya164 yes they look correct to me as a start, but I would mention that you also want to add a call to actually jetify in the postinstall

```suggestion
  Then ensure jetifier runs after all package installs by adding a call to `npx jetify` in the `postinstall` lifecycle script of your `package.json` file, and make sure you have run `npx jetify` at least once prior to attempting to build for android.


  Next, add these two lines in `gradle.properties` under `android` folder
```

Or it might be better to just link to the jetifier docs? That would be https://github.com/mikehardy/jetifier#to-jetify--convert-node_modules-dependencies-to-androidx

satya164: @mikehardy we do mention it below:

Screenshot 2019-09-30 at 5 18 55 PM

I'd emphasise that this part of the setup instructions is for **React Native 0.59 and lower** which didn't include `androidx`. That's why I'm not sure about this part.

mikehardy: It is actually very difficult to be prescriptive for jetifier. It is possible to use jetifier in the forward or reverse direction on RN59 and it depends on what the project needs. Necessary complexity to be flexible unfortunately - the full 2x2 matrix of "library is AndroidX or support libraries / project is AndroidX or support libraries" is handled. So I would counsel to just expose the information "this library has converted to AndroidX, if you need special handling in your project for either AndroidX or Android support libraries, please consult the relevant jetifier documentation" or similar

A bit vague, sorry

Rizwanjamal: @satya164 But I think the instructions for setting up `jetifier` mentioned in the Doc for **React Native 0.59 and lower** isn't working without adding these config values inside gradle.properties these days.
image

mikehardy: but, that's only for people that want to *go forward* to AndroidX. It is vital for that. If people still want to go reverse, those are the exact wrong things to put there as they are part of an AndroidX transition (they tell the gradle plugin to do the jetify transformation to AndroidX on all static artifacts like jars zips and aars)

Rizwanjamal: @mikehardy Got it. So why are making it necessary to configure `jetifier`. Is it necessary ?
image

because, it's creating problem in making build, & that's the reason we need to set these config values... Can we add these as an optional configuration for only case, if user get the exact error ?

mikehardy: It depends? If you are using RN59 *and* (a critical *and*) you want to go to AndroidX then you need those in gradle config, and you need a call to `npx jetify`. So that sounds like your case. But maybe you are using RN59 and you want to stay on support libraries. Then you *must not* have those, and you need a call to `npx jetify -r`. And that is the use case for other people. It depends.

Rizwanjamal: @mikehardy Alright !

satya164:
Closing as per the discussion.

PR #530, opened 2019-10-01
Fix Deep linking iOS docs using deprecated method

hkellaway:
The Deep linking iOS docs indicate to use the long deprecated `appliation:openURL:sourceApplication:annotation:` method. Update to use preferred `application:openURL:options:` method.


hkellaway:
@satya164 done!

PR #548, opened 2019-10-14
Update some url(s) in the docs from *.html to *.md

kam89:
to re-link the url to *.md instead of *.html.
to ensure the url in the docs are browse-able.


kam89:
> Can you post a screenshot showing that this works with docusaurus?

Hi, 

Please find the screenshot below for existing url link.
This is pointing to **navigation-lifecycle.html**
Screenshot from 2019-10-15 07-32-33-crop

And this resulted the below if we click the url in the page:
Screenshot from 2019-10-15 07-32-42

If we change the url to the .md, then it will redirect to the github.com as per below:
Screenshot from 2019-10-15 07-33-28


But need to take note that if this docs is going to be pull over for the hosting for https://reactnavigation.org/ then this changes is not working for the hosting if the hosting folder contains the _*.html_ files instead of the _*.md_ files.

This changes is worked for **GitHub** only.

PR #550, opened 2019-10-16
Fix transitionConfig types link

sgtpepper43:
The link explaining `TransitionConfig` is broken, so I've updated it to point to the type definitions on the ~~v1.9.4 tag~~ `1.0` branch. Since it's pointing to the ~~tag~~ branch instead of master it shouldn't break again. The transition config is gone in v5 so there's nothing to change in the original file.


sgtpepper43:
For sure, yeah pointing to the branch would be better.

PR #553, opened 2019-10-18
Fix broken link for pull request template in contributing.md

JaredClayborn:
# READ ME PLEASE!

Thanks for opening a PR!

I had a terrible brain fart and had to fix my mistake.  The previously versioned files (1.x, 2.x, etc) don't need to be changed when I make a new version, that doesn't make sense.  So those were reverted, that's why there are a ton of commits for this pull request

Mtoivation:
I was hoping to contribute to one of these repos during Hacktoberfest and found a broken link in the contributing.md.  This could be (is?) the first of several pull requests during Hacktoberfest.

Test plan:
I ran checked the link locally

Code formatting:
I didn't change any formatting, just fixed the broken link


satya164:
Thanks for the PR

PR #554, opened 2019-10-21
Fix missing tabBarIcon option key on Tab navigation sample

zetavg:
Reference: https://reactnavigation.org/docs/en/next/tab-based-navigation.html

### DIFF
--- a/website/versioned_docs/version-5.x/tab-based-navigation.md
+++ b/website/versioned_docs/version-5.x/tab-based-navigation.md
@@ -1,0 +1,265 @@
+---
+id: tab-based-navigation
+title: Tab navigation
+sidebar_label: Tab navigation
+---
+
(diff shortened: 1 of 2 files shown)


website/versioned_docs/version-5.x/tab-based-navigation.md
```diff
@@ -0,0 +1,265 @@
+---
```
satya164: we don't have a version-5.x folder

satya164:
Closing due to no response

website/versioned_docs/version-5.x/tab-based-navigation.md
zetavg: I'm following the instructions on https://github.com/react-navigation/react-navigation.github.io/blob/source/.github/PULL_REQUEST_TEMPLATE.md to add versioned docs since this is for `5.x`. Should this file not to be add? Or should it be add to a different location?

PR #555, opened 2019-10-23
update(screen-tracking) change analytics module example

jamsch:
The repository for `react-native-google-analytics-bridge` has now been archived and users are now recommended to use `@react-native-firebase/analytics`. This commit changes the example to use the more relevant library.


jamsch:
Updated

PR #557, opened 2019-10-27
Update screen-options-resolution.md

anhtuank7c:
Add example code for  **stack contains a tab navigator which contains another stack**


satya164:
Thanks for the PR, but we need more details in the guide, similar to https://reactnavigation.org/docs/en/navigation-options-resolution.html

anhtuank7c:
@satya164 I got it. I will do it when I have free time.

WoLewicki:
This is already done in "Setting parent screen options based on child navigator's state", am I right @anhtuank7c ? If so, I would close the PR.

anhtuank7c:
@WoLewicki true. We can close this PR

PR #561, opened 2019-10-30
Update redux-integration.md

Gguigre:
I think that 'connected' was transformed into `connect`ed.

This might have happened with a search/replace operation on the word "connect", transformed into "`connect`", that would have taken the connected word.

This change was made in : https://github.com/react-navigation/react-navigation.github.io/commit/a0c0567991679b04464a93ed21a12e1a30977705


satya164:
Thanks for the PR. It's intentional. It's using the `connect` function as verb.

PR #568, opened 2019-11-05
blog: add guide on using react-navigation 5 with ui-kitten

artyorsh:
Hello 👋 

I was talking for a while to @satya164 about writing this article. We at Akveo were inspired with such global changes in React Navigation, so we decided to write a step-by-step guide on using it with our library [UI Kitten](https://github.com/akveo/react-native-ui-kitten). And we hope you might be interested in it too :)

In this article I've described the usage of every Navigator bound to our navigation components: 

- Stack Navigator
- Top Tab Navigator
- Bottom Tab Navigator
- Drawer Navigator

I've also added a short note on using it with TypeScript.

You might also be interested in [demo app](https://github.com/artyorsh/react-navigation-ex-demo/tree/complete-exmaples) that is built on the latest `@react-navigation/*` packages (We're going to maintain this repo till v5 stable is released to fit all breaking changes).

Current package versions:

- `@react-navigation/core@5.0.0-alpha.23`
- `@react-navigation/native@5.0.0-alpha.16`
- `@react-navigation/stack@5.0.0-alpha.37`
- `@react-navigation/material-top-tabs@5.0.0-alpha.18`
- `@react-navigation/bottom-tabs@5.0.0-alpha.22`
- `@react-navigation/drawer@5.0.0-alpha.23`

The image links used in the article are currently replaced with relative paths for the easier review.

**EDIT** 

I wanted to create a draft pull request. Unfortunately, I've picked another option while opening.
I believe we can keep this open and have a discussion about this article here :)


website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
+---
+title: Binding React Navigation 5 to UI Kitten
+author: Artur Yorsh
+authorURL: https://twitter.com/artyorsh
+---
+
+The complete step-by-step guide on using new React Navigation API.
```
satya164: I'd add that this is guest post and not from the React Navigation team.

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+
+export const AuthNavigator = (): React.ReactElement => (
+  <Stack.Navigator headerMode='none'>
+    <Stack.Screen name='Sign In' component={SignInScreen}/>
+    <Stack.Screen name='Sign Up' component={SignUpScreen}/>
+  </Stack.Navigator>
+);
+```
+
+The syntax is simple. You just import `createXNavigator` from the navigator package of your choice and use `Navigator` and `Screen` components from the value returned by this function.
```
satya164: It's better to avoid words like "simple" and "just"

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+
+```bash
+git clone https://github.com/artyorsh/react-navigation-ex-demo
+```
+
+### Step 1. Authentication flow
+
+Assuming that your app users will need to authorize before getting to the home screen, we will need to create both Authentication and Home navigators. Then we're going to combine it with simple stack navigation and pick the initial screen depending on the user authorization status.
+
+Open `./src/navigators/auth.navigator.tsx` file and paste the following code:
```
satya164: It'll be nice to have file names link to the respective files in the GH repo so you can click them to see the file

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+
+
+```js
+import React from 'react';
+import { createStackNavigator } from '@react-navigation/stack';
+import { SignInScreen, SignUpScreen, ResetPasswordScreen } from '@app-scenes/auth';
```
satya164: I'd avoid using imports like `@app-scenes/auth` for example code and use relative imports. Otherwise it's not obvious which file we're importing just by looking at the code.

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+import { createStackNavigator } from '@react-navigation/stack';
+import { AuthNavigator } from './auth.navigator';
+import { AppRoute } from './app-routes';
+
+const Stack = createStackNavigator();
+
+export const AppNavigator = (config): React.ReactElement => (
+  <NavigationNativeContainer>
+    <Stack.Navigator headerMode='none'>
+      <Stack.Screen name={AppRoute.AUTH} component={AuthNavigator}/>
```
satya164: Will be better to avoid a wrapper navigator. In our docs regarding auth, we encourage using conditionals instead of navigators for auth https://reactnavigation.org/docs/en/next/auth-flow.html

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+    <Stack.Screen name={AppRoute.SIGN_IN} component={SignInScreen}/>
+    <Stack.Screen name={AppRoute.SIGN_UP} component={SignUpScreen}/>
+    <Stack.Screen name={AppRoute.RESET_PASSWORD} component={ResetPasswordScreen}/>
+  </Stack.Navigator>
+);
+```
+
+In this example, we're using a `createStackNavigator` function to create simple stack navigation between Sign In, Sign Up and Reset Password screens. Under `Stack Navigator` we mean the default navigation behavior between screens: with slide-from-right animation on iOS, and slide-in-top on Android.
+
+In app.navigator.tsx file and replace the placeholder screen with Auth Navigator. This will make authentication screens to be the starter point of your app.
```
satya164:
```suggestion
In `app.navigator.tsx` file and replace the placeholder screen with Auth Navigator. This will make authentication screens to be the starter point of your app.
```

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+
+Say, our app has both in-progress and finished tasks. So, you should separate them to avoid a mess. Here you can make it with two tabs on the home screen. To do this, we need to have three screens: two for tabs and one master screen for navigation management between tabs. Unlike the Stack Navigator component, the Top Tabs `Navigator` has a special prop for the component to control navigation between tabs - `tabBar`. We will use it to configure the tab bar with UI Kitten components.
+
+Open `./src/navigators/todo.navigator.tsx` file and paste the following code:
+
+```js
+import React from 'react';
+import { createStackNavigator } from '@react-navigation/stack';
+import { createMaterialTopTabNavigator } from '@react-navigation/material-top-tabs';
+import { TodoScreen, TodoInProgressScreen, TodoDoneScreen } from '@app-scenes/todo';
```
satya164: I think naming the tab bar component as `TodoScreen` is a bit confusing as I'd expect it to be a screen, not a tab bar.

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+import { AppRoute } from './app-routes';
+
+const Stack = createStackNavigator();
+const TopTab = createMaterialTopTabNavigator();
+
+export const TodoNavigator = (): React.ReactElement => (
+  <TopTab.Navigator tabBar={TodoScreen}>
```
satya164: `tabBar` accepts a render callback, not component:

```suggestion
  <TopTab.Navigator tabBar={props => <TodoScreen {...props} />}>
```

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+  const onTabSelect = (index: number): void => {
+    const { [index]: selectedTabRoute } = props.state.routeNames;
+    props.navigation.navigate(selectedTabRoute);
+  };
+
+  return (
+    <SafeAreaLayout insets={SaveAreaInset.TOP}>
+      <Toolbar title='React Navigation Ex 🐱'/>
+      <TabBar selectedIndex={props.state.index} onSelect={onTabSelect}>
+        <Tab icon={GridIcon} title='IN PROGRESS'/>
```
satya164: Instead of hardcoding the tab items, you should do something like this:

```js
state.routes.map(route => {
  const { options } = descriptors[route.key];

  return <Tab key={route.key} icon={options.icon} title={options.title} />
});
```

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+import { TabBar, Tab, Divider } from 'react-native-ui-kitten';
+import { SafeAreaLayout, SaveAreaInset, SafeAreaLayoutElement } from '@app-components/safe-area-layout.component';
+import { Toolbar } from '@app-components/toolbar.component';
+import { DoneAllIcon, GridIcon } from '@app-assets/icons';
+
+export const TodoScreen = (props): SafeAreaLayoutElement => {
+
+  const onTabSelect = (index: number): void => {
+    const { [index]: selectedTabRoute } = props.state.routeNames;
+    props.navigation.navigate(selectedTabRoute);
```
satya164: Instead of navigating manually, you can use the `onTabPress` prop passed to the tab bar. This makes sure events such as `tabPress` etc. get triggered.

```js
onTabPress({ route: state.routes[index] });
```

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+      </TabBar>
+      <Divider/>
+    </SafeAreaLayout>
+  );
+};
+```
+
+With the code above we render `TabBar` component with two tabs inside: one per each screen inside `TodoNavigator`. Then, we use the React Navigation state to pass `selectedIndex` and `onSelect` props to navigate between screens. So, when the user taps one of the tabs, the `TabBar` component calls `onTabSelect` function and this is the place where we need to navigate to the corresponding route.
+
+Finally, open `app.navigator.tsx` file and add the `TodoNavigator as a Home screen. Now you're able to navigate the home screen with todo tabs after sign in.
```
satya164:
```suggestion
Finally, open `app.navigator.tsx` file and add the `TodoNavigator` as a Home screen. Now you're able to navigate the home screen with todo tabs after sign in.
```

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,429 @@
...
+import React from 'react';
+import { NavigationNativeContainer } from '@react-navigation/native';
+import { createStackNavigator } from '@react-navigation/stack';
+import { AuthNavigator } from './auth.navigator';
+import { TodoNavigator } from './todo.navigator';
+import { AppRoute } from './app-routes';
+
+const Stack = createStackNavigator();
+
+export const AppNavigator = (config): React.ReactElement => (
```
satya164: I'd call it `props` rather than `config`

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
artyorsh: I believe "conditional navigators" is a bit complex example for this guide. I've tried to use Auth screens to demonstrate stack navigation between `Reset Password` and `Sign Up` / `Sign In` screens. Would it be better to refer `auth-flow` guide instead of creating complete example regarding auth?

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
artyorsh: That is exactly what it does, because `TodoScreen` [is a function](
https://github.com/artyorsh/react-navigation-ex-demo/blob/complete-exmaples/src/scenes/todo/todo.component.tsx#L26) :) Just inline syntax

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
artyorsh: Thanks for the explanation 👍 
Unfortunately, I've tried using this function, but it throws ts errors since it accepts the `route` and `preventDefault` function. See complete log on a [screenshot](https://monosnap.com/file/NOyY03FoyiO0E5QEUR8xLPNYsCwnqg)

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
artyorsh: Correct. Also, a drawer menu could be refactored with something like this, but I find this way more intuitive. Can we leave this as it is?

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,420 @@
...
+Just like Top/Bottom tab navigators, the drawer navigator also has a special property for declaring custom drawer view. Use a `drawerContent` property to pass the custom view to the navigator. Open [./src/scenes/home/drawer-home.component.tsx](https://github.com/artyorsh/react-navigation-ex-demo/blob/complete-exmaples/src/scenes/home/drawer-home.component.tsx) file and add the following code:
+
+```js
+import React from 'react';
+import { Drawer, DrawerElement, MenuItemType } from 'react-native-ui-kitten';
+import { AppRoute } from '../../navigation/app-routes';
+import { SafeAreaLayout, SaveAreaInset} from '../../components/safe-area-layout.component';
+import { InfoIcon, LogoutIcon } from '../../assets/icons';
+
+const drawerData: MenuItemType[] = [
```
satya164: Instead of hardcoding `drawerData`, le's do this:

```js
// Show items for each screen in the navigator
const drawerData = state.routes(route => {
  const options = descriptors[route.key];

  return {
    routeName: route.name,
    title: options.title,
    icon: options.icon,
  }
});

// Add extra menu item for a logout button
drawerData.push({
  title: 'Log Out',
  icon: LogoutIcon,
});
```

Then user can specify `title` and `icon` in screen's `options` prop

website/blog/2019-11-04-binding-react-navigation-5-to-ui-kitten.md
```diff
@@ -0,0 +1,420 @@
...
+];
+
+export const DrawerHomeScreen = (props): DrawerElement => {
+
+  const onMenuItemSelect = (index: number): void => {
+    const { [index]: selectedItem } = drawerData;
+
+    switch (selectedItem.title) {
+      case 'Log Out':
+        props.navigation.navigate(AppRoute.AUTH);
```
satya164: For stuff like logging out, instead of navigate, we recommend to do the action that actually logs the user out, such as removing the auth token from the store. If you're conditionally rendering screens for auth like in the docs, then everything is automatically handled.

Note that even if you use the navigate approach, you still need to clear the token even in a basic app. So it's extra work.

PR #571, opened 2019-11-08
[V4] fix themes related examples

Simek:
This PR fixes small issues in themes related examples :
* missing `theme` variable initialization in static context usage example
* missing context initialization in _"Themes inside `navigationOptions`"_
* missing `ThemeContext` import in _"Theming tabs and other similar navigators"_

### DIFF
--- a/website/versioned_docs/version-4.x/themes.md
+++ b/website/versioned_docs/version-4.x/themes.md
@@ -112,6 +112,7 @@ class MyButton extends React.Component {
   static contextType = ThemeContext;
 
   render() {
+    const theme = this.context;
     return (
       <TouchableOpacity
         style={{ backgroundColor: theme === 'light' ? '#000' : '#fff' }}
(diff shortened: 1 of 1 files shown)


WoLewicki:
Thanks for contribution! I think we should add `ThemeConstants` to "Themes inside navigationOptions" to be consistent and don't import `ThemeContext` in "Theming tabs and other similar navigators" but rather create it as in "Themes inside navigationOptions". Could you do it? And what do you think @satya164 ?

satya164:
User needs to import `ThemeContext` to be able to use the theme from React Navigation. Creating own theme context means it's a different theme system. I think user shouldn't create separate context if they just need to integrate with React Navigation's themes.

WoLewicki:
Yes, but in the examples, there are used custom values from `ThemeConstants`, like this 
`let currentTheme = ThemeConstants[screenProps.theme];`

satya164:
I'm not following. I mean current `ThemeConstants` usage is ok, but the `ThemeContext` should be imported instead of being created by the user.

WoLewicki:
Ok, so just change the creation of ThemeContext to importing it from `react-navigation` ?

satya164:
Yeah

Simek:
@WoLewicki @satya164 Thank you for the comments. 

I have modified the PR according to your discussion. I hope I have understood the required change properly.

WoLewicki:
Thanks for the update! I think it is okay now, if @satya164 thinks so too, we can merge it.

PR #573, opened 2019-11-13
Add docs about updating parent navigator options based on child navigator state

satya164:


docs/screen-options-resolution.md
```diff
@@ -70,14 +70,150 @@ export default function App() {
...
+       <Tab.Navigator>
+         <Tab.Screen name="Feed" component={FeedScreen} />
+         <Tab.Screen name="Profile" component={ProfileScreen} />
+         <Tab.Screen name="Account" component={AccountScreen} />
+       </Tab.Navigator>
+     );
+   }
+   ```
+
+So what's happening here? The `route` prop contains a `state` property which refers to the child navigator's state (in this case it's the tab navigator since that's what we're rendering). We are getting the value of the currently active route name from this state and setting appropriate title for the header.
```
osdnk:
```suggestion
So what's happening here? The `route` prop contains a `state` property which refers to the child navigator's state (in this case it's the tab navigator since that's what we're rendering). We are getting the value of the currently active route name from this state and setting an appropriate title for the header.
```

docs/screen-options-resolution.md
```diff
@@ -70,14 +70,150 @@ export default function App() {
...
+
+> Note: The `route.state` property may not exist at all. This will always happen if we have never navigated inside the tab navigator. So it's very important to handle this case, otherwise your app will crash.
```
osdnk:
```suggestion
> Note: The `route.state` property may not exist at all. This will always happen if we have never navigated inside the tab navigator. So it's very important to handle this case, otherwise, your app will crash.
```

PR #576, opened 2019-11-16
Important note - active and inactiveTintColor

JohnPremKumar:
Currently the activeTintColor and inactiveTintColor works only if the color attribute is not mentioned in the label style object.
This is not properly conveyed in the current documentation

# READ ME PLEASE!

Thanks for opening a PR!


WoLewicki:
Thanks for the contribution! I added a note in labelStyle about it. Check #604.

PR #577, opened 2019-11-17
[Docs]: Fix Typo in Code Example

medelman17:
This PR fixes the following typo: 

Before: 

`export const LocalizationContext = React,createContext();`

After: 

`export const LocalizationContext = React.createContext();`

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -36,7 +36,7 @@ console.log(Localization.locale);
 Next let's store our `locale` in the state of our root app component and then thread it through `LocalizationContext` to make it available throughout our app.
 
 ```js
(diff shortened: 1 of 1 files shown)


WoLewicki:
Thanks for your contribution! It is fixed now.

WoLewicki:
Hi,
I would just like to say that we are aware that the way this PR was ended was unprofessional and we promise to make every effort to avoid such situations in the future. At this point, I would like to emphasize how important external contributions are for us.
As an apology and an expression of concern for our users, we would like to offer you a few React Navigation and Software Mansion stickers.

PR #579, opened 2019-11-26
[Docs] Add NavigationNativeContainer references around Stack Navigators

CatalinVoss:
Don't we still need these? Certainly seems to fail without them.


WoLewicki:
You are right. Thanks for the contribution! I will fix it and add snacks for these examples.

PR #585, opened 2019-12-10
Fixed icons in example 'Tab navigation' - Add badges to icons

gianpaj:
https://reactnavigation.org/docs/en/tab-based-navigation.html#add-badges-to-icons

Hee's the snack as well: https://snack.expo.io/rJOD3FT6H


satya164:
What does this fix?

gianpaj:
this fixes the icons on the bottom navigation bar on the Snack

see the web preview here
https://snack.expo.io/@react-navigation/tabs-with-icons-v3

and then here
https://snack.expo.io/rJOD3FT6H

satya164:
the snack is broken, but code in the documentation doesn't have any issues

PR #589, opened 2019-12-20
useNavigation is in '@react-navigation/core' of ver. 'next'

yzalvov:
Otherwise, import from **/native** leads to an error as `useNavigation` there is `undefined`.


satya164:
Please update to latest alpha.

PR #592, opened 2019-12-31
Fix wrong import path for SafeAreaView

perryfaro:
Fix wrong import path for SafeAreaView


WoLewicki:
Thanks for the PR! According to https://github.com/react-native-community/react-native-safe-area-view, the current code seems to be alright. Are you sure it is necessary?

perryfaro:
Thanks for your reply, as far as I can see the docs for version 5 (next) don’t mention to install that package only the react-native-safe-area-context. If you than follow the docs you will get an error on the SafeAreaView import.

WoLewicki:
Yes, you are right. There should be only react-native-safe-area-context needed. Thanks for contribution!

PR #595, opened 2020-01-05
Fix: import

kopax:
Imports was broken, it seems that the correct packge is react-navigation-hooes, see https://github.com/react-navigation/native/issues/37

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/docs/connecting-navigation-prop.md
+++ b/docs/connecting-navigation-prop.md
@@ -31,7 +31,7 @@ Alternatively, you can use the `useNavigation` to provide the `navigation` prop 
 ```js
 import * as React from 'react';
 import { Button } from 'react-native';
-import { useNavigation } from '@react-navigation/native';
+import { useNavigation } from 'react-navigation-hooks';
 
 function GoToButton({ screenName }) {
   const navigation = useNavigation();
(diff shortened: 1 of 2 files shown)


satya164:
The documentation is for React Navigation 5 (next). You're using React Navigation 4. Please read the docs for the stable version at https://reactnavigation.org/

kopax:
Thanks for enlightening me, what should I do now? Should I use react-navigation 5? I am just starting and haven't implemented anything that I can't throw.

satya164:
It's upto you if you want to use the stable version it the prerelease version.

kopax:
I will rather be on the prerelease version if I can do the same work, but since I have no clue I stick to the last stable release.

Are you working on it? Because I am not. I have no clue if I can use it for production. (documentation? features status?), maybe you can give me a hint?

satya164:
It's fairly stable with very minor issues. But I see that you need web support which isn't implemented yet.

kopax:
Actually, for web I am using react-router. I have started to implement with v4, but I will definitly switch to v5 alpha if you say so. I will study the next documentation, in case you have time to answer this little question, is there an upgrade guide I can follow or is the API for the following the same?

What's is the biggest difference between the two?

I am basically just using `createStackNavigator`, `createSwitchNavigator` and `createAppContainer`,.

satya164:
https://blog.expo.io/announcing-react-navigation-5-0-bd9e5d45569e

PR #602, opened 2020-01-08
Update native stack navigator guide

WoLewicki:
Fixed missing NavigationNativeContainer in examples and added snacks to them

### DIFF
--- a/docs/native-stack-navigator.md
+++ b/docs/native-stack-navigator.md
@@ -28,19 +28,24 @@ enableScreens();
 
 To use this navigator, import it from `@react-navigation/native-stack`:
 
+<samp id="simple-native-stack">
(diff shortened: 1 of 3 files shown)


satya164:
We don't have the container in individual navigator examples because user should have it at the root (it's in getting started guide)

WoLewicki:
But if the function is called App, then we treat it as the root in examples, yes?

satya164:
We can change it to not be called APP

WoLewicki:
Should I change it and remove NavigationNativeContainer or leave it as it is?

satya164:
@WoLewicki let's change it

WoLewicki:
Ok, I will change it.

PR #609, opened 2020-01-11
fix: Add react-navigation-stack to expo install command

hassan-bazzi:
Otherwise, this guide doesn't work: https://reactnavigation.org/docs/en/hello-react-navigation.html

# READ ME PLEASE!

Thanks for opening a PR!


WoLewicki:
Thanks for PR! As far as I know it is stated on the https://reactnavigation.org/docs/en/hello-react-navigation.html that you have to install react-navigation-stack first to be able to use it. Isn't it what you requested?
image

PR #610, opened 2020-01-14
chore: add filter language and version on search

eriveltonelias:
When you want to search something on the docs, this may have same results but for different versions, even if you're looking v4 or next version. This PR should handle this situation, if you're looking at v5 the search results will be only for v5 and the language as well.


satya164:
Thanks a lot!

eriveltonelias:
😁

satya164:
Hey, seems this broke search for some reason. I have reverted it for now. Would be awesome if you could take a look when possible.

eriveltonelias:
for sure, I'll check it.

PR #613, opened 2020-01-20
fix: correct example code for useNavigationState

halileohalilei:
I was going through the docs for the useNavigationState hook and realized that the example code contained the wrong signature.


satya164:
Thanks!

PR #615, opened 2020-01-22
Add navigationRef

DLevai94:
Fix an error on React Navigation V5 docs


WoLewicki:
Thanks for your contribution! You are right, it should be there.

PR #617, opened 2020-01-23
feat: add snack to use-navigation-state

WoLewicki:
Added snack to useNavigationState and fixed typo.


website/static/examples/next/use-navigation-state.js
```diff
@@ -0,0 +1,89 @@
...
+function useIsFirstRouteInParent() {
+  const route = useRoute();
+  const isFirstRouteInParent = useNavigationState(
+    state => state.routes[0].key === route.key
+  );
+
+  return isFirstRouteInParent;
+}
+
+function getPreviousRouteName() {
```
satya164:
```suggestion
function usePreviousRouteName() {
```

PR #635, opened 2020-02-05
fix broken link

bell-steven:
Fixed a broken link.


bell-steven:
I left the /en/ part of the URL, because otherwise it redirects to /en/ but discards #navigating-to-a-screen-in-a-nested-navigator.

WoLewicki:
Thanks for your contribution! You are right, merging it.

PR #636, opened 2020-02-05
Docs: embed tree diagram in modal.md

schellack:
This change is to embed the tree diagram image directly into modal.md so that the reader doesn't have to click through to see it. This is useful as the diagram is mentioned more than once in this document to help visualize where a modal fits in when using multiple, nested navigators.


WoLewicki:
Thanks for your contribution! I don't know if this image should be shown on the page. Wdyt @satya164 ?

satya164:
I think it makes sense. Thanks for the PR!

PR #637, opened 2020-02-06
Fixing small typo on blog post from 2020-02-06

stigi:
Thanks for the great writeup! So hyped about react-navigation 5!

This PR only contains changes to the blog, and not to documentation.


osdnk:
<3

PR #657, opened 2020-02-07
fix: docs edit url

eriveltonelias:
currently when we try to edit the docs from the `Edit button` is redirecting to a wrong path.


WoLewicki:
Thanks for your contribution! You are right, there should be `/` at the end.

eriveltonelias:
thanks @WoLewicki

PR #660, opened 2020-02-11
feat: add deep linking guide

WoLewicki:
Add a blogpost/guide concerning basic functionalities of deep linking in v5. The version with the guide is the latest commit and blogpost is a second latest commit.


osdnk:
Can we make it a guide instead?

PR #668, opened 2020-02-18
docs: useLinking ref is missing

terrierscript:
In deep-linking documentation, `ref` is declared with `const ref = React.useRef();`
https://reactnavigation.org/docs/en/deep-linking.html 

But in useLinking doc, it's may lacking


WoLewicki:
Thanks for your contribution! You are right.

PR #672, opened 2020-02-19
docs: add TextInput to import

pamccombs:
It's obvious after reading the error message but is still very helpful for walking through error free.

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/website/versioned_docs/version-5.x/params.md
+++ b/website/versioned_docs/version-5.x/params.md
@@ -107,7 +107,8 @@ function CreatePostScreen({ navigation, route }) {
 
   return (
     <>
-      <TextInput
+      {/* Add {TextInput} to react-native import */}
+      <TextInput 
         multiline
         placeholder="What's on your mind?"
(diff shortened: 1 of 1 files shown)


pamccombs:
My apologies, this is my first try at contribution. I'm unsure where or how exactly to place the semantics.

WoLewicki:
Thanks for your contribution! I think that adding such note in the code makes it less readable. It is better. It would be better if you deleted these notes and just added `import { TextInput } from 'react-native'` at the top of this snippet. Can you do it?

pamccombs:
I've recreated a separate pull request with your suggestion. Thank you

--
Patrick "Reese" McCombs
Software Engineer

pamccombs:
I've actually edited this one instead and have removed the other request.

WoLewicki:
Nice! If only you could add an empty line after the import to follow the convention and then it is ready to be merged.

pamccombs:
I have created a separate pull request with the correct changes.

PR #693, opened 2020-02-26
feat: improve the npm/yarn installation instruction

eriveltonelias:
**Before**:

Screen Shot 2020-02-26 at 09 09 52

**After**: 

Screen Shot 2020-02-26 at 09 09 34

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -116,6 +116,7 @@ module.exports = {
           sidebarPath: require.resolve('./sidebars.js'),
           editUrl:
             'https://github.com/react-navigation/react-navigation.github.io/edit/source/',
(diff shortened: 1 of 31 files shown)


eriveltonelias:
> `--save` is not necessary when running `npm` as it adds the dependency to `package.json` like `yarn` for a while now. Can you remove the addition of `--save`?

Yes, done :)

slorber:
FYE we may add this plugin to docusaurus repo: https://github.com/facebook/docusaurus/pull/3469

eriveltonelias:
will be amazing @slorber, because right now we're using a copy/paste code from docusaurus core. 

I just subscribed for this plugin PR and definitely, I want to import that from docusaurus when is available to use.

PR #710, opened 2020-03-04
chore: add announcement migration v4 to v5 on Home page

eriveltonelias:
react-navigation 5 is really cool! 😎 so it will be cool if when users open our website they can see the migration guide easily.


eriveltonelias:
@satya164 I'll let you decide this one.

Issue #719, opened 2020-03-14
Feature request: add a functional version of the code samples


jay-jlm:
With the increasing adoption of functional components and React hooks (including now its support in  recent versions of the Expo SDK) it would be great if the there's option to view a functional version of the sample code in the documentation.
More and more projects no longer use classes and the conversion can troublesome specially for less experienced developers.

PR #721, opened 2020-03-17
docs: update ref example with function component

Naturalclar:
# READ ME PLEASE!

Thanks for opening a PR!

This issue is related to #719 

updated app-container example from class component to function component.


jarvisluong:
Thank you for making a PR! However, I believe instead of completely change the class-based component example, we should have both versions to remain neutral of which implementation method is preferred.

eriveltonelias:
Thanks, @Naturalclar for your PR, but unfortunately I don't think we should take this approach ( by replacing the class to hooks ) :( 
as @jarvisluong said, will be nice to have both... but it will require extra work to keep both examples up to date and we need to figure out a simple way to show them.

PR #734, opened 2020-03-29
docs: Update auth-flow.md

JB1905:


satya164:
Thanks for the PR. But this is intentional to also allow the token to be `undefined`.

PR #738, opened 2020-04-01
Add note regarding Stack transition config

Rijk:
I had to browse through multiple source files to find out this information. Hopefully this saves someone a couple hours in the future.

References: https://github.com/react-navigation/react-navigation/blob/3df65e28197db3bb8371059146546d57661c5ba3/packages/stack/src/views/Stack/CardStack.tsx#L468

### DIFF
--- a/versioned_docs/version-5.x/stack-navigator.md
+++ b/versioned_docs/version-5.x/stack-navigator.md
@@ -548,14 +548,14 @@ Stack Navigator exposes various options to configure the transition animation wh
   />
   ```
 
-- `cardStyleInterpolator` - This is a function which specifies interpolated styles for various parts of the card. It is expected to return at least empty object, possibly containing interpolated styles for container, the card itself, overlay and shadow. Supported properties are:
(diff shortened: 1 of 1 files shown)


versioned_docs/version-5.x/stack-navigator.md
```diff
@@ -566,6 +566,8 @@ Stack Navigator exposes various options to configure the transition animation wh
   - `layouts` - Layout measurements for various items we use for animation.
     - `screen` - Layout of the whole screen. Contains `height` and `width` properties.
 
+  > **Note that when a screen is not the last, it will use the next screen's transition config.** This is because many transitions involve an animation of the previous screen, and so these two transitions need to be kept together to prevent running two different kinds of transitions on the two screens (for example a slide and a modal). You can check the `next` parameter to find out if you want to animate out the previous screen.
```
jarvisluong, line 569:
```suggestion
  > **Note that when a screen is not the last, it will use the next screen's transition config.** This is because many transitions involve an animation of the previous screen, and so these two transitions need to be kept together to prevent running two different kinds of transitions on the two screens (for example a slide and a modal). You can check the `next` parameter to find out if you want to animate out the previous screen. For more information about this parameter, see [Animation](/docs/stack-navigator#animations) section.
```

jarvisluong:
Thanks!

PR #740, opened 2020-04-02
fix: rewrite nested navigators bubble example

GoudekettingRM:
#### Rewritten the nested navigators bubble example

**[Example of nested navigators](https://reactnavigation.org/docs/nesting-navigators/) this PR applies to.**

In the [old example](https://reactnavigation.org/docs/nesting-navigators/#navigation-actions-are-handled-by-current-navigator-and-bubble-up-if-couldnt-be-handled), it is said that if `navigate("Settings")` is called from the `Profile` screen, the nested Stack Navigator will handle it. However, there is no nested navigator for the `Profile` screen in the example. Furthermore, it states that if `navigate("Home")` is called from the `Profile` screen, it will bubble up and the parent Tab Navigator will handle it. There is no parent Tab Navigator, however. Therefore, there will be no bubbling up through the navigators. 

The new example reflects what is actually happening with the nested navigators. If `navigate('Messages')` is called from the `Feed` screen, which is in the same Navigator, the navigate will be handled locally, by the Home component's Tab Navigator.
Calling navigate('Settings') from the Feed screen will then bubble up to the Stack Navigator, since the Home component's Tab Navigator does not have a 'Settings' screen.

### DIFF
--- a/versioned_docs/version-5.x/nesting-navigators.md
+++ b/versioned_docs/version-5.x/nesting-navigators.md
@@ -50,7 +50,7 @@ For example, when you press the back button inside a nested stack navigator, it'
 
 ### Navigation actions are handled by current navigator and bubble up if couldn't be handled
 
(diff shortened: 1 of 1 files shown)


jarvisluong:
LGTM!

PR #753, opened 2020-04-17
Adding Types to documentation and Typescript Example

jarvisluong:
Linked issue: https://github.com/react-navigation/react-navigation.github.io/issues/751

Basically I think Typescript example are really not much different from JS version, just some special case when we need to explicitly set the type. I'm thinking to add the types info to the documentation, similar to React native website's style

Screenshot 2020-04-17 at 4 41 15 PM


jarvisluong:
My plan is to first go through the navigators and add the type info, then look at some example to see if any of them need a dedicated typescript version

eriveltonelias:
Hey @jarvisluong, what is the status of this? can you what Satya suggested and close this one, please?

PR #769, opened 2020-05-06
docs: better understanding for "screenOptions"

dougmbarcellos:
The "options prop" part was confusing.


eriveltonelias:
Hey @dougmbarcellos! thank you so much for your PR! but we already moved the `options` to another page.

PR #777, opened 2020-05-13
Add section about Universal Links

LinusU:
As discussed in https://github.com/react-navigation/react-navigation/pull/8252


LinusU:
Thanks for the quick pull! ❤️

PR #779, opened 2020-05-15
docs: Update navigation prop setOptions example

leimonio:
Not entirely sure if this is intended, but I believe in this example use of `React.useLayoutEffect` is misleading.
In the example component no specific DOM manipulations/calculations are made, so `React.useEffect` should be just enough for this case. 

I've tried using `React.useEffect` and it works fine for updating the header title.


satya164:
`useEffect` is async and runs some time after the screen has rendered. You normally want to update the title as soon as the screen has rendered, not after some time. This is why `useLayoutEffect` is used in the example.

PR #781, opened 2020-05-16
Add information about HeaderBackButton

Naturalclar:
I've ran into a scene where I'd like to override the onPress of back button without changing its looks.
I've searched and found that you can import `HeaderBackButton` component, but I couldn't find that being mentioned in the document, so I'm adding information for other people who runs into same problem.

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-5.x/header-buttons.md
+++ b/versioned_docs/version-5.x/header-buttons.md
@@ -87,6 +87,8 @@ The back button will be rendered automatically in a stack navigator whenever it 
 
 Generally, this is what you want. But it's possible that in some circumstances that you want to customize the back button more than you can through the options mentioned above, in which case you can set the `headerLeft` option to a React Element that will be rendered, just as we did with `headerRight`. Alternatively, the `headerLeft` option also accepts a React Component, which can be used, for example, for overriding the onPress behavior of the back button. Read more about this in the [api reference](/docs/stack-navigator/#headerleft).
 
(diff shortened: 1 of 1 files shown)


eriveltonelias:
@Naturalclar can you update the link of `HeaderBackButton`, please?

versioned_docs/version-5.x/header-buttons.md
```diff
@@ -87,6 +87,8 @@ The back button will be rendered automatically in a stack navigator whenever it
 
 Generally, this is what you want. But it's possible that in some circumstances that you want to customize the back button more than you can through the options mentioned above, in which case you can set the `headerLeft` option to a React Element that will be rendered, just as we did with `headerRight`. Alternatively, the `headerLeft` option also accepts a React Component, which can be used, for example, for overriding the onPress behavior of the back button. Read more about this in the [api reference](/docs/stack-navigator/#headerleft).
 
+If you would like to retain the view of back button and only override the `onPress` method, you can import [HeaderBackButton](https://github.com/react-navigation/react-navigation/blob/master/packages/stack/src/views/Header/HeaderBackButton.tsx) from `@react-navigation/stack` and assign that component to the `headerLeft` option.
```
Naturalclar, line 90:
```suggestion
If you would like to retain the view of back button and only override the `onPress` method, you can import [HeaderBackButton](https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/views/Header/HeaderBackButton.tsx) from `@react-navigation/stack` and assign that component to the `headerLeft` option.
```

Naturalclar:
@eriveltonelias thanks! I've updated the link

eriveltonelias:
thanks!

PR #787, opened 2020-05-18
Navigation type for deeply nested navigators

JavanPoirier:
I personally encountered a type error on the second property of the navigation call. I assumed a type would be defined in the library for allowing these additional parameters on deeply nested navigators. There does not appear to be, so I have added a use case and how to define this type yourself.

Additions only apply to v5 docs.


satya164:
Thanks for the PR. We now have a built-in type for this, so I'm closing the PR.

https://reactnavigation.org/docs/typescript/#type-checking-screens-and-params-in-nested-navigator

PR #814, opened 2020-06-25
Fix docs for InitialParams

RaphaelBRodrigues:
It should be 'route.params.user' instead of route.params.me in the session The navigation prop ( https://reactnavigation.org/docs/upgrading-from-4.x/ 

Before:
stack
function

Now:
now


RaphaelBRodrigues:
Eu que agradeço @eriveltonelias

PR #828, opened 2020-07-14
docs: changes in usage of header

anjushaKhandavalli:
Added documentation for the changes in the usage of header

# READ ME PLEASE!

Thanks for opening a PR!


anjushaKhandavalli:
Updated the semantic pull request

anjushaKhandavalli:
Using the header: null to hide the header in V5 leading to crashes. This got fixed by changing the header to a function that returns an empty component.

eriveltonelias:
Hi @anjushaKhandavalli, thanks for your PR! 

on v5 you can use the `headerShown` prop to achieve this goal, for more info check the docs please https://reactnavigation.org/docs/stack-navigator/#headershown

anjushaKhandavalli:
Yeah yes. But header: null is leading to crashes in V5

satya164:
> Yeah yes. But header: null is leading to crashes in V5..which is allowed in V4

With React Navigation v4 and latest version of Stack Navigator, even if it doesn't crash, there's a deprecation warning if you use `header: null`.

> this is not mentioned anywhere in the document about header not accepting null in V5

If you're upgrading from older versions of navigators rather than their latest compatible versions in v4, you should first take a look at the v4 upgrade guide. This is already mentioned there.

https://reactnavigation.org/docs/4.x/upgrading-from-3.x#upgrading-packages

PR #845, opened 2020-08-10
docs: Update 'Passing params to a previous screen'

jaworek:
I'm currently developing app that passes params from subsequent screens. One screen also has a screen that passes back additional route params.
I was not sure if I needed to pass all params to new screen or if the behaviour is the same as ```setParams```.
It turns out that second behaviour is present. 
I think it will be useful to mention it in the documentation.

### DIFF
--- a/versioned_docs/version-5.x/params.md
+++ b/versioned_docs/version-5.x/params.md
@@ -146,6 +146,32 @@ navigation.navigate('Account', {
 
 See [Nesting navigators](nesting-navigators.md) for more details on nesting.
 
+## Behavior of params
+
+On every action (e.g. ```navigation```, ```setParams```, etc.) params are getting merged. 
+If your previous screen already contains route params, new params will added or overwrite current.
+
+```
+// old params
+{
+  user: 'jane',
+  age: 20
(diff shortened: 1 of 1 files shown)


satya164:
Params are merged for every action, not just `setParams` or `navigate`. So I'm not sure mentioning it specifically in this example makes sense. If it wasn't obvious, maybe the better place to mention it is the getting started guides.

jaworek:
Sure. I couldn't find this information anywhere in documentation and had to test it in my app what is the actual behavior.
I think it would be nice to mention it somewhere. Please, let me know where you think is the best place, so I can move it there.

satya164:
Thanks, I think we can add a section in this page about what happens when passing params to an existing screen: https://reactnavigation.org/docs/params

jaworek:
Great! Thank you :D

jaworek:
I added new section. I don't like the title, so if you have any other ideas, let me know.

satya164:
There have been a lot of changes regarding this in 6.x and 7.x, so for now, I'm closing this. We can revisit this later.

PR #855, opened 2020-08-29
Update screen-tracking.md

MohamadKh75:
Updates due to https://github.com/invertase/react-native-firebase/pull/4145


satya164 (approved):
Thanks!

Would be great if you can also send a PR for v5.

PR #857, opened 2020-08-31
Docs: Update custom-android-back-button-handling.md

andrekovac:
Add `BottomTabNavigator` to list of navigators which can be intercepted to add custom Android back button behavior.


satya164 (changes requested):
Thanks for the PR. I think this section is outdated. Custom back button handling should work for all navigators, not just stack navigator and bottom tab navigator.

PR #861, opened 2020-09-09
Update modal.md

Billydubb:
The example shown in the documentation was wrong. The parent stack navigator needs to have its `headerMode` prop set to `none`. The `headerShown` option is irrelevant for this example.

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-5.x/modal.md
+++ b/versioned_docs/version-5.x/modal.md
@@ -62,11 +62,10 @@ function MainStackScreen() {
 
 function RootStackScreen() {
   return (
-    <RootStack.Navigator mode="modal">
+    <RootStack.Navigator mode="modal" headerMode="none">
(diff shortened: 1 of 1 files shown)


satya164:
Why is it wrong? The example demonstrates that you have to hide the header only for the screen containing a nested stack. The other screen can still have a header.

Billydubb:
I thought this was about opening an full-screen modal, without any header, when navigating to the screen that says "This is a modal", just as it is shown in the gif at the top of the page.

https://reactnavigation.org/assets/images/modal-demo-c21e394139257ff16390d46aa6ba5bc2.gif

satya164:
The GIF needs to be updated. The examples are written this way intentionally. So I'm closing this PR.

PR #867, opened 2020-09-25
Replace DrawerItems with DrawerNavigationItems for 4.x

ansonhoyt:
When upgrading from 3.x to 4.x, I updated the packages and imports per https://reactnavigation.org/docs/4.x/upgrading-from-3.x/ and everything worked except the drawer nav. When I used `DrawerItems`, I got an error:
```
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

My suggested change fixed my little Drawer Navigator. I think maybe this component was renamed in `react-navigation-drawer` 1.4.0, which is the version the migration guide recommends for 4.x.

This might be the wrong change. I saw the 2.4.0 [changelog](https://github.com/react-navigation/react-navigation/blob/4.x/packages/drawer/CHANGELOG.md#240-2020-02-24) mentions DrawerItems is re-exposed. In that case, perhaps the migration guide should be updated to recommend installing this newer version?

I lost motivation tracing the source with repos moving around, but hopefully this PR is somewhat helpful.

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-4.x/drawer-navigator.md
+++ b/versioned_docs/version-4.x/drawer-navigator.md
@@ -42,7 +42,7 @@ The route configs object is a mapping from route name to a route config, which t
 - `gestureHandlerProps` - Props to pass to the underlying pan gesture handler.
 - `lazy` - Whether the screens should render the first time they are accessed. Defaults to `true`. Set it to `false` if you want to render all screens on initial render.
 - `unmountInactiveRoutes` - Whether a screen should be unmounted when navigating away from it. Defaults to `false`.
-- `contentComponent` - Component used to render the content of the drawer, for example, navigation items. Receives the `navigation` prop and `drawerOpenProgress` for the drawer. Defaults to `DrawerItems`. For more information, see below.
+- `contentComponent` - Component used to render the content of the drawer, for example, navigation items. Receives the `navigation` prop and `drawerOpenProgress` for the drawer. Defaults to `DrawerNavigationItems`. For more information, see below.
 - `contentOptions` - Configure the drawer content, see below.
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR. However both are exported in latest versions of drawer so it shouldn't cause any issues: https://github.com/react-navigation/react-navigation/blob/10c6c3280f07949bce0725531d10eee21d9ad5fb/packages/drawer/src/index.tsx#L18

You likely had an older version or outdated cache.

ansonhoyt:
Yes, I did have an older version.

I was following [Upgrading from 3.x](https://reactnavigation.org/docs/4.x/upgrading-from-3.x/) guide, and used the provided command, `yarn add react-navigation react-navigation-stack@^1.7.3 react-navigation-tabs@^1.2.0 react-navigation-drawer@^1.4.0`, which doesn't install the 2.x.x version needed to work. Ideally these docs should work, and a major version jump beyond the documented one was unexpected, but it is a small issue 🤷 😄 

So, the migration guide is slightly mismatched to the working code. I should've suggested a migration docs PR instead.

Thank you.

satya164:
> I was following Upgrading from 3.x guide, and used the provided command, yarn add react-navigation react-navigation-stack@^1.7.3 react-navigation-tabs@^1.2.0 react-navigation-drawer@^1.4.0, which doesn't install the 2.x.x version needed to work.

If you're installing older versions, then it means that you don't need to change any code. The migration guide for navigators are only for when you upgrade navigators (which is in separate upgrading your navigators section).

PR #868, opened 2020-09-25
remove deprecated react-native-header-scroll-view

Simek:
This small PR removes the link to the deprecated and unmaintained [`react-native-header-scroll-view`](https://github.com/jonsamp/react-native-header-scroll-view) library and the related example link, which also leads to 404.


eriveltonelias (approved):
thanks!

PR #869, opened 2020-09-28
Update getting started to note about Windows

alexanderblackh:
Now that react-native windows supports react-navigation, I've added a couple of notes in the getting started page due to the fact that react-native-gesture-handler breaks the app on Windows and prevents a successful build.

### DIFF
--- a/versioned_docs/version-5.x/getting-started.md
+++ b/versioned_docs/version-5.x/getting-started.md
@@ -51,7 +51,7 @@ npm install react-native-reanimated react-native-gesture-handler react-native-sc
 
 > Note: You might get warnings related to peer dependencies after installation. They are usually caused by incorrect version ranges specified in some packages. You can safely ignore most warnings as long as your app builds.
 
-> Note: If using this project with [react-native-windows](https://github.com/microsoft/react-native-windows), omit`react-native-gesture-handler`.
+> Note: If using this project with [react-native-windows](https://github.com/microsoft/react-native-windows), omit `react-native-gesture-handler`.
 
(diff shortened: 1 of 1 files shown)


versioned_docs/version-5.x/getting-started.md
```diff
@@ -65,7 +67,7 @@ To finalize installation of `react-native-gesture-handler`, add the following at
 import 'react-native-gesture-handler';
 ```
 
-> Note: If you skip this step, your app may crash in production even if it works fine in development.
+> Note: Unless you are using react-native-windows, do not skip this step as your app may crash in production even if it works fine in development.
```
satya164, line 70:
```suggestion
> Note: If you are building for Android or iOS, do not skip this step, or your app may crash in production even if it works fine in development. This is not applicable to other platforms.
```

versioned_docs/version-5.x/getting-started.md
```diff
@@ -51,6 +51,8 @@ npm install react-native-reanimated react-native-gesture-handler react-native-sc
 
 > Note: You might get warnings related to peer dependencies after installation. They are usually caused by incorrect version ranges specified in some packages. You can safely ignore most warnings as long as your app builds.
 
+> Note: If using this project with [react-native-windows](https://github.com/microsoft/react-native-windows), omit`react-native-gesture-handler`.
```
satya164, line 54:
```suggestion
> Note: If using this project with [react-native-windows](https://github.com/microsoft/react-native-windows), omit `react-native-gesture-handler`.
```

satya164 (approved):
Thanks!

PR #870, opened 2020-10-01
fix: mention useCallback to passing additional props

Gregoirevda:
As the children prop is a function, it seems like using `useCallback` is more appropriate.

Example:
```js
import React, {useCallback} from 'react';

function App() {
const [counter, setCounter] = useState(0);

const renderScreen = useCallback(({router, navigation}) => {
  return <HomeScreen  router={router} navigation={navigation} counter={counter} />
}, [counter])

  return <Stack.Navigator>
    <Stack.Screen name="Home">
      {renderScreen}
    </Stack.Screen>
 </Stack.Navigator>
}
```

With `useMemo` or `PureComponent`, I'm a lever deeper (rendering the component), so I would need to handle not only my custom props, but `router`, `navigation`, ... props coming from React Navigation.

With `useCallback`, I just need to pass my additional props!

ref: https://reactjs.org/docs/hooks-reference.html#usecallback

If you agree with this, I could **replace** `useMemo` and `PureComponent` with `useCallback` and class methods and probably add an example.
I'll would that for all docs versions then

### DIFF
--- a/versioned_docs/version-5.x/hello-react-navigation.md
+++ b/versioned_docs/version-5.x/hello-react-navigation.md
@@ -133,7 +133,7 @@ Sometimes we might want to pass additional props to a screen. We can do that wit
    </Stack.Screen>
    ```
 
-> Note: By default, React Navigation applies optimizations to screen components to prevent unnecessary renders. Using a render callback removes those optimizations. So if you use a render callback, you'll need to ensure that you use [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) or [`React.PureComponent`](https://reactjs.org/docs/react-api.html#reactpurecomponent) for your screen components to avoid performance issues.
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR. Wrapping your component with `React.memo`/`React.PureComponent` is the proper way instead of using `useCallback`. We want to memoize the content of the component, not the callback itself.

> so I would need to handle not only my custom props, but router, navigation, ... props coming from React Navigation.

I'm not sure what you mean by that. You don't need to handle anything with `React.memo` or `React.PureComponent`. e.g. - `React.memo(MyComponent)` or `class MyComponent extends React.PureComponent` doesn't need any manual dependency tracking.

> With useCallback, I just need to pass my additional props!

Your component will still re-render unnecessarily because memoizing the function doesn't gurrantee that the function won't be called multiple times.

PR #871, opened 2020-10-06
docs: remove deprecated accessibilityStates prop

Naturalclar:
# READ ME PLEASE!

Changed `accessibilityStates` prop to `accessibilityState`, as `accessibilityStates` prop is no longer being used in react-native

ref: https://github.com/facebook/react-native/pull/26168

```
Thanks for opening a PR!

```

I've looked at version-3.x docs and the changed content did not exist in the older docs.


eriveltonelias (approved):
thanks!

PR #881, opened 2020-10-21
Update Reactiflux invite URL

vcarl:
The old invite was putting joiners into an error state due to the new Code of Conduct gate! They'd see a "messages failed to load" error, and they'd need to navigate to #welcome first to accept. Not ideal.


satya164:
Thanks!

PR #882, opened 2020-10-22
beforeRemove will not work when tab is unfocused

NishargShah:
In the react-navigation top and bottom tabs, beforeRemove not works, so I described here. Followed https://github.com/react-navigation/react-navigation/issues/8851


satya164:
Thanks!

PR #887, opened 2020-11-04
Update material-bottom-tab-navigator.md

Ripplz:
The `theme` prop doesn't have any effect on this component's tab bar label font as it does react-native-paper's `BottomNavigation` component, so after digging through the source code, I found that you can pass a fully customized React component as this prop instead of just a string.

# READ ME PLEASE!

Thanks for opening a PR!


satya164:
What do you mean by "The `theme` prop doesn't have any effect on this component's tab bar label font as it does react-native-paper's `BottomNavigation` component"?

satya164:
Thanks for the PR, however, the types don't allow to pass anything other than a string. While passing a react element works right now it's not supported and can break in the future.

PR #897, opened 2020-11-25
Fix hyperlinks to github react-navigation repo in version 4.x docs

Kamalnrf:
The hyperlinks were pointing to the archived `react-navigation` repo so updated them to active repo and that particular branch file.


satya164 (approved):
Thanks!

PR #902, opened 2020-12-03
Add a small tip for Firebase Dynamic Links

Cattari:
In case of usage of Firebase Dynamic links, getting the initial URL should be performed by firebase itself


satya164:
Thanks for the PR. We now have a full example with Firebase dynamic links: https://reactnavigation.org/docs/deep-linking/#third-party-integrations

So I'm closing this.

PR #905, opened 2020-12-10
Update sponsors

handipriyono:
Update sponsors


brentvatne:
hiya! sorry we didn't see this. i updated it in https://github.com/react-navigation/react-navigation.github.io/commit/2ed4e87b6e87dafc4ebb2280c4f6fdf978f24ff7

PR #913, opened 2020-12-30
unexpected inline code syntax removed

kamal-hossain:
There was an unexpected inline code in `versioned_docs/version-5.x/redux-integration.md`.


johnjago:
It's like this throughout the page, probably to make it clear that it is referring to the `connect` function.

```js
connect(state => ({ value: state.count }))
```

I think this is okay.

eriveltonelias (approved):
Thanks!

PR #919, opened 2021-01-07
feat: Hide Snack query params on URL bar

IjzerenHein:
# Why

Hides the query parameters from the URL bar when opening a Snack example. This is made possible by the new `hideQueryParams=true` option https://github.com/expo/snack/blob/main/docs/url-query-parameters.md

**before**

image

**after**

image

 # How

- Add `'&hideQueryParams=true` to url


IjzerenHein:
Hmm, preview doesn't seem to work. Will investigate.. 🧐
https://deploy-preview-919--react-navigation-docs.netlify.app

eriveltonelias (approved):
Thanks ❤️

PR #926, opened 2021-01-27
Update navigation-state.md

wave-light:
Correct the property name to plural: routeNames


eriveltonelias:
Thanks for your PR! but we already merged this change https://github.com/react-navigation/react-navigation.github.io/pull/917

PR #927, opened 2021-01-27
updated themes.md to include appearances URL

faris-mohamed10:
# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-5.x/themes.md
+++ b/versioned_docs/version-5.x/themes.md
@@ -79,7 +79,7 @@ import { DefaultTheme, DarkTheme } from '@react-navigation/native';
 
 ## Using the operating system preferences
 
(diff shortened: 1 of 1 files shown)


versioned_docs/version-5.x/themes.md
```diff
@@ -79,7 +79,7 @@ import { DefaultTheme, DarkTheme } from '@react-navigation/native';
 
 ## Using the operating system preferences
 
-At the time of writing, `react-native` does not currently support detecting the operating system color scheme preferences in the core ([you can follow this pull request](https://github.com/facebook/react-native/pull/26172)). Until it is part of core and you have updated to the version that includes it, you can use `react-native-appearance`.
+appearance is a part of react native and can see more details here https://reactnative.dev/docs/appearance.
```
eriveltonelias, line 82: I think we should explain a little bit more about using the system preferences ( Appearance module ) before we reference this to a link

eriveltonelias (changes requested):
thanks for your PR! I just left one small comment.

PR #930, opened 2021-02-12
fix: Update getting-started.md for drawer swipe gesture to work on Android

akmal578:
# Read Me Please

To fix drawer navigation not responding to the swipe on android, MainActivity.java need to be updated to make the gesture handlers work properly. Thus, I recommended putting the instructions in the docs for easier initial setup and saving much more time.

### DIFF
--- a/versioned_docs/version-5.x/getting-started.md
+++ b/versioned_docs/version-5.x/getting-started.md
@@ -57,7 +57,8 @@ From React Native 0.60 and higher, [linking is automatic](https://github.com/rea
 
 ### Updating MainActivity.java
 
-Location:
+Files location:
+
 > ..\MyApp\android\app\src\main\java\com\MyApp\MainActivity.java
 
 React Native 0.60 migrated from Support Library to AndroidX. React Native Gesture Handler is not yet compatible with AndroidX.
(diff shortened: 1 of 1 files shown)


jakub-gonet:
AFAIK android changes aren't necessary when you use hoc. Do you happen to use gesture handler 1.10.0?

akmal578:
> AFAIK android changes aren't necessary when you use hoc. Do you happen to use gesture handler 1.10.0?

Yes I'm currently using gesture handlers 1.10.0 . Well if you think its ok then maybe we can close it

jakub-gonet:
Try upgrading to 1.10.1 and check if there's still a problem without using native setup.

We had misnamed file responsible for hoc in rngh, and metro was prioritizing non-Android specific one (it would choose x.ts over x.android.tsx due to the metro matching ts files first and not finding '.android' alternative). It was resolved in the 1.10.1.

akmal578:
> Try upgrading to 1.10.1 and check if there's still a problem without using native setup.
> 
> We had misnamed file responsible for hoc in rngh, and metro was prioritizing non-Android specific one (it would choose x.ts over x.android.tsx due to the metro matching ts files first and not finding '.android' alternative). It was resolved in the 1.10.1.

Already tried upgrading it and it seems working fine even without the native setup.  Thanks

PR #934, opened 2021-02-16
fix: grammar error in latest navigation events documentation

ryantk:
This is a fix to a simple grammar error I spotted when reading the documentation.


eriveltonelias (approved):
thanks!

PR #936, opened 2021-02-19
route.params error

maksshilov:
- options={({ route }) => ({ title: route.params.name })}
+ options={({ route }) => ({ title: route.name })}

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-5.x/headers.md
+++ b/versioned_docs/version-5.x/headers.md
@@ -44,7 +44,7 @@ function StackScreen() {
       <Stack.Screen
         name="Profile"
         component={ProfileScreen}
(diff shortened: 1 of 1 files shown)


versioned_docs/version-5.x/headers.md
```diff
@@ -44,7 +44,7 @@ function StackScreen() {
       <Stack.Screen
         name="Profile"
         component={ProfileScreen}
-        options={({ route }) => ({ title: route.params.name })}
```
eriveltonelias, line 47: Hey @maksshilov, thanks for your PR! 

this is the expected behavior for this example, we want to show how to change the `title` based on `params` prop.

PR #940, opened 2021-02-23
docs: update react-native-masked-view link

dcangulo:
Summary:

Migrated docs from `@react-native-community/masked-view` to `@react-native-masked-view/masked-view`


eriveltonelias (approved):
thanks!

LcTwisk:
I think this might result in some issues because is the code there are still imports to the old library (or at least the old name) 🤔 
[Here](https://github.com/react-navigation/react-navigation/blob/7b353a4aeabe204284b358af204caebe1db093c1/packages/stack/src/views/MaskedViewNative.tsx#L7) for example.

It also looks like the new library won't be used until the next major update: https://github.com/react-navigation/react-navigation/pull/9283

PR #941, opened 2021-02-24
Update preventing-going-back.md

zjkuang:
I found an issue that in some cases the same listener may be added repetitively causing multiple alerts. To fix that, perform a removeListener() before addListener().


satya164:
Why would there be an extra listener added automatically that you need to remove? Also you're passing an anonymous function to `removeListener`, so how would it know which listener to remove?

Also you need to actually return the value of `addListener` so that event can be cleaned up.

PR #951, opened 2021-03-17
fix: Remove reference to createTabNavigator()

benomatis:
Remove reference to createTabNavigator() - it's been removed in `react-navigation@3.x`


eriveltonelias (approved):
thanks!

PR #977, opened 2021-04-21
Fix invalid link in documentation for stack-navigator

patrickgalbraith:
The headerLeft option documentation has an invalid link and lists the wrong property names.


eriveltonelias (approved):
Thanks!

PR #989, opened 2021-05-07
docs: update custom android back handling code to match RN docs

Phoenix1355:
Updated the code snippet for custom android back handling to match the [functional component example for BackHandling](https://reactnative.dev/docs/backhandler) in the React Native documentation. It's a simple change but could mitigate potential mistakes with string typos.


satya164:
Thanks for the PR! However this seems to be fixed in the docs already so I'm closing this.

PR #995, opened 2021-05-16
docs: Coinbase uses React Navigation

JonnyBurger:
I extracted the bundle of Coinbase and saw it uses React Navigation. To not have Coinbase in there twice, I replaced Coinbase Pro with the main Coinbase app since it’s way more popular.


satya164 (approved):
Thank you

PR #1013, opened 2021-06-19
Correct screen tracking call to expo-firebase-analytics API.

### LINKED ISSUE #1012: Implementation of expo-firebase-analytics seems wrong
I think that the example of screen tracking using `expo-firebase-analytics` in the v5 and v6 is incorrect.

It references `analytics()` instead of the `Analytics` that is imported from `expo-firebase-analytics` at the top of the file, which looks to me like a leftover `@react-native-firebase/analytics` implementation from v4 of the docs.

https://github.com/react-navigation/react-navigation.github.io/blob/2f11f169907d3210679453b9fb6fd608cbfd76b8/versioned_docs/version-6.x/screen-tracking.md#L41-L48

Compare this to how the Expo docs for `expo-firebase-analytics` implement screen tracking with React Navigation [here](https://docs.expo.io/versions/latest/sdk/firebase-analytics/#react-navigation) that uses `Analytics.setCurrentScreen(currentScreen)`.

jimpala:
Fixes #1012 (which appears to apply for the v5 and v6 docs).

### DIFF
--- a/versioned_docs/version-5.x/screen-tracking.md
+++ b/versioned_docs/version-5.x/screen-tracking.md
@@ -40,10 +40,7 @@ export default () => {
           // The line below uses the expo-firebase-analytics tracker
           // https://docs.expo.io/versions/latest/sdk/firebase-analytics/
           // Change this line to use another Mobile analytics SDK
-          await analytics().logScreenView({
-            screen_name: currentRouteName,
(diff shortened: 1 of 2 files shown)


eriveltonelias (approved):
thanks!

PR #1020, opened 2021-07-14
docs: Adding troubleshooting info

denissb:
### Motivation
When trying out the `next` version of `@react-navigation/stack` the `NavigationContainer` fails to mount with an error `"TypeError: Cannot read property 'SafeAreaProviderCompat' of undefined"`

There is a comment here: https://github.com/react-navigation/react-navigation/blob/main/packages/elements/src/MaskedViewNative.tsx#L17, but it can take some time to find the cause of the error.

Instead it is better to mention the error in the troubleshooting guide for the people who want to try out the new version on their existing projects.

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-6.x/troubleshooting.md
+++ b/versioned_docs/version-6.x/troubleshooting.md
@@ -194,6 +194,16 @@ cd ..
 
 Now rebuild the app and test on your device or simulator.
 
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR, however, this doesn't look like a common problem, but a bug. Can you open an issue in React Navigation repo with a repro so we can fix it?

PR #1023, opened 2021-07-22
Argument to `createURL` is now a blank string

phatmann:
If you use use `createURL('/')`. you will get a triple-slashed prefix in a standalone app; e.g. "myapp:///". You have to use `createURL('')`, which gives a double-slashed prefix in a standalone app; e.g. "myapp://". A blank string also produces a correct prefix in an Expo-hosted app.

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-5.x/deep-linking.md
+++ b/versioned_docs/version-5.x/deep-linking.md
@@ -37,7 +37,7 @@ import * as Linking from 'expo-linking';
 // you are using older versions, you can upgrade or use Linking.makeUrl instead,
 // but note that your deep links in standalone apps will be in the format
 // scheme:/// rather than scheme:// if you use makeUrl.
-const prefix = Linking.createURL('');
+const prefix = Linking.createURL();
 
 function App() {
   const linking = {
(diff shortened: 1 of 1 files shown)


versioned_docs/version-5.x/deep-linking.md
```diff
@@ -37,7 +37,7 @@ import * as Linking from 'expo-linking';
 // you are using older versions, you can upgrade or use Linking.makeUrl instead,
 // but note that your deep links in standalone apps will be in the format
 // scheme:/// rather than scheme:// if you use makeUrl.
-const prefix = Linking.createURL('/');
+const prefix = Linking.createURL('');
```
brentvatne, line 40:
```suggestion
const prefix = Linking.createURL();
```

brentvatne (approved):

satya164:
Thanks for the PR, I believe this is now handled in React Navigation so it'll work fine. AFAIK there is some issue with empty string too but I'm unclear since this API has changed.

Let me know if this is still an issue.

PR #1029, opened 2021-08-03
Update algolia to search 6.x docs by default

rrebase:
### Description

Update algolia to search 6.x docs by default as that's now the latest version.


satya164 (approved):
Thanks

PR #1039, opened 2021-08-13
fix(custom-navigators): incorrect typing for Props

JoniVR:
I was getting Typescript errors in my custom navigator following an upgrade from a beta version of v6 to `6.0.2`, so I decided to look into the codebase of React Navigation itself. Seems to me like the docs still require updating here.

See [here](https://github.com/react-navigation/react-navigation/blob/2084fb859ffa7c8ce2de859dba777a1227ff05c2/packages/bottom-tabs/src/navigators/createBottomTabNavigator.tsx#L21) for example in `createBottomTabNavigator`


satya164 (approved):
Thank you

PR #1043, opened 2021-08-23
fix: Missing installations for Drawer Based Navigation

efefurkankarakaya:
In the docs there are 2 related Drawer pages and one of them does not contain required packages or routing the related installation package. 

Page 1: https://reactnavigation.org/docs/drawer-based-navigation
Page 2: https://reactnavigation.org/docs/drawer-navigator/


satya164 (changes requested):
Thanks for the PR, but the installation instructions differ depending on whether you use expo, and whether you use reanimated 1 or 2. So instead of duplicating it here, I'd rather link to the drawer navigator's installation section and emphasize it.

kacperkapusciak:
I'm closing this PR because this issue was fixed in 1c5de7d

PR #1052, opened 2021-09-02
docs: update mocking requirements for v6 navigators

wwdrew:
The v6 documentation details for mocking are slightly out of date, it's only required to mock react-native-reanimated and react-native-gesture-handler when using certain navigators.


satya164 (approved):
Thanks for the PR

PR #1060, opened 2021-09-14
correct HeaderBackButton usage

kidroca:
Fixes incorrect usage example


kidroca:
Something to note about JSX code highlighting 

When the usage example contains `jsx` the code block should be defined using 

    ```jsx
    ...
    ```

and not just `js`

    ```js
    ...
    ```

See the difference: 
image

satya164:
Thanks for the PR. However this seems to be fixed already so I'm closing this

PR #1068, opened 2021-09-29
docs: update pop function document

sridhar-sp:
update pop function document


satya164:
Thanks for the PR. However, the API is documented on the action's page. I don't think it makes sense to document only `pop`'s arguments here. So I'm closing this.

PR #1076, opened 2021-10-09
Update used-by.md

vuducnamcs:
Hello, We are Tomaho(er) (an enterprise software company) use React Navigation.

# READ ME PLEASE!

Thanks for opening a PR!


eriveltonelias (changes requested):
Screen Shot 2021-10-31 at 10 33 42

Hey @vuducnamcs, looks like your app link is not displaying as expected, can you fix that please?

PR #1082, opened 2021-10-20
Missing `pod install` step for installation

Miraj98:
Need to `pod install` after installing `react-native-pager-view`

### DIFF
--- a/versioned_docs/version-6.x/material-top-tab-navigator.md
+++ b/versioned_docs/version-6.x/material-top-tab-navigator.md
@@ -34,6 +34,7 @@ If you have a bare React Native project, in your project directory, run:
 
 ```bash npm2yarn
 npm install react-native-pager-view
+cd ios && pod install
 ```
 
 ## API Definition
--- a/versioned_docs/version-6.x/material-top-tab-navigator.md
+++ b/versioned_docs/version-6.x/material-top-tab-navigator.md
@@ -34,8 +34,8 @@ If you have a bare React Native project, in your project directory, run:
 
 ```bash npm2yarn
 npm install react-native-pager-view
-cd ios && pod install
 ```
+> If you are developing for IOS, You'd need to run: **pod install**
 
 ## API Definition
 


versioned_docs/version-6.x/material-top-tab-navigator.md
```diff
@@ -34,6 +34,7 @@ If you have a bare React Native project, in your project directory, run:
 
 ```bash npm2yarn
 npm install react-native-pager-view
+cd ios && pod install
```
react-one, line 37: That is specific to only ios. And the documentation already tackled that in the [starting guide](https://reactnavigation.org/docs/getting-started/#installing-dependencies-into-a-bare-react-native-project). And [troubleshooting guide](https://reactnavigation.org/docs/troubleshooting/#im-getting-an-error-null-is-not-an-object-evaluating-rngesturehandlermoduledefaultdirection).  And one need to start with the starting guide.

A possibility is to add a small alert that reminds of not forgetting to execute `pod install`. But then that need to be done for all other packages. And navigators. It's better to open an issue, and open a discussion about this.

Miraj98: The starting guide does not require you to install `react-native-pager-view` . Its a dependency specifically required for material-top-tab-navigator. I personally faced this issue while setting up this the main navigator library and if the docs had simply mentioned to run `pod install` after adding `react-native-pager-view`

react-one: I can understand, i'm personally an advocate of having information straight in the place, even when you do repeat, as fast access is always productive. The way it can go there, would be with an info container that say `if you are developing for IOS, don't forget to run: pod install`.

Miraj98: Have made the changes now. Hope this works :)

satya164 (approved):
Thanks for the PR

PR #1034, opened 2021-08-06
Fixing yarn autoconverter for broken `--save-dev` flag

pvinis:
`yarn add --save-dev X` is wrong, `yarn add --dev X` is right.

before:
Screenshot 2021-08-06 at 21 41 58

after is weirdly still the same? maybe netlify is not rebuilding or something? on node repl, it works fine.

PR #1088, opened 2021-11-04
fix: `npm2yarn` conversion of `--save-dev` install option

mmmoussa:
On https://reactnavigation.org/docs/devtools/#useflipper the command `npm install --save-dev react-native-flipper` is being converted to `yarn add --save-dev react-native-flipper` which is wrong because `--save-dev` is invalid in yarn and should instead be `--dev`.


mmmoussa:
Closing since there's already #1034

Issue #1095, opened 2021-12-02
`IsSigned` in instead of `IsSignedIn` in auth-flow page


rafaeltab:
# Current Behavior
In the section `Don't manually navigate when conditionally rendering screens​` when referring to the `IsSignedIn` variable it was referred to as ``` `IsSigned` in   ```

# Expected Behavior
It should be referred to as ``` `IsSignedIn` ```

# Versions
| Version | Bug present |
|---------|-------------|
| 1.x     | ❌           |
| 2.x     | ❌           |
| 3.x     | ❌           |
| 4.x     | ❌           |
| 5.x     | ✔           |
| 6.x     | ✔           |

eriveltonelias:
closing in favor of https://github.com/react-navigation/react-navigation.github.io/pull/1096

PR #1096, opened 2021-12-02
docs: `IsSigned` in instead of `IsSignedIn` in auth-flow page

rafaeltab:
Fix issue #1095 
Only involves docs.


eriveltonelias (approved):
Thanks!

PR #1101, opened 2021-12-23
Fix grammar in v5 and v6 of params.md

sandypockets:
# READ ME PLEASE!

Thanks for opening a PR!

--- 

Hey there,

I am following along in some of the React Navigation docs, and noticed a grammar error. The `params.md` file of Docs version 5 and 6 were using `an user id` instead of `a user id`. This PR fixes that grammar in both of those files. It was not present in earlier versions. 

Files changed: 

* `/versioned_docs/version-4.x/params.md`
* `/versioned_docs/version-4.x/params.md`

Thanks!


eriveltonelias (approved):
Thanks!

PR #1105, opened 2022-01-12
docs: mention what example app would be displaying for listed JS state

rctay:
# READ ME PLEASE!

Thanks for opening a PR!

___

Hi there, this change further fleshes out which route an app is displayed and its corresponding navigation state.. For example, it might not be immediately obvious to a newcomer that among the 2 routes of `Home` and `Settings`, it would be `Settings` that is being displayed and not `Home`.


satya164:
Thanks for the PR. The intention here is not to understand how the state object is displayed, but just to show how it maybe structured. How it is displayed would depend on the navigator. For example, a navigator may decide to display both screens side by side.

If we describe what it may show, I'd prefer to do it after explaining what each property is, e.g. in the state object that's shown after the descriptions.

So I'm closing this for now. Feel free to add a description to the later state object that explains what it may show.

PR #1106, opened 2022-01-12
docs: mention index is always last route in a stack

rctay:
# READ ME PLEASE!

Thanks for opening a PR!

___

Hi there, this change make the docs mention this, in case a developer tries to specify a value for index
apart that is not equal to the last route in a stack and sees that it has no effect, eg. issue #10056.


satya164 (changes requested):
Thanks for the PR. I think adding "For example..." to the existing paragraph feels a bit out of place. I'd create a new paragraph for this.

However, the exact structure of navigation state is described in a separate document and it also talks about invalid states & corrections (under "Partial state objects"), so I feel like it'd be better to link there directly instead of adding this here. People need to be familiar with state object to use `reset` anyway, so that page is a pre-requisite.

https://reactnavigation.org/docs/navigation-state/

rctay:
Thanks for the review, and thanks for pointing me to that section, the new text does fit in better there.

rctay:
Hi @satya164,  

> Thanks for the PR. I think adding "For example..." to the existing paragraph feels a bit out of place. I'd create a new paragraph for this.
> 
> However, the exact structure of navigation state is described in a separate document and it also talks about invalid states & corrections (under "Partial state objects"), so I feel like it'd be better to link there directly instead of adding this here. People need to be familiar with state object to use `reset` anyway, so that page is a pre-requisite.
> 
> https://reactnavigation.org/docs/navigation-state/

I've added the paragraph to navigation state under the Partial state objects section, how does this look to you?

satya164 (approved):
Thanks for the PR

PR #1108, opened 2022-01-15
fix: `tabBar` component example

tiagohermano:
the current example is passing a wrong type of argument to the `navigate` method.


satya164:
Thanks for the PR, however the current code is correct and this change is incorrect.

PR #1114, opened 2022-02-11
Update configuring-links.md

denisk20:
Use proper id value.

# READ ME PLEASE!

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-6.x/configuring-links.md
+++ b/versioned_docs/version-6.x/configuring-links.md
@@ -469,7 +469,7 @@ const state = {
           { name: 'Feed' },
           {
             name: 'Profile',
-            params: { id: 'jane' },
+            params: { id: '42' },
           },
         ],
       },


kacperkapusciak:
Hi @denisk20 :wave:
Thanks for submitting a PR! :tada: 

String ID values are ubiquitous in programming eg. UUID. Also, an arbitrary string (as here `jane`) is just an example and it's definitely a proper ID value.

Cheers

denisk20:
@kacperkapusciak the point of the PR was for this ID to match the one from [here](https://github.com/react-navigation/react-navigation.github.io/pull/1114/files#diff-20df68d576bf20f54be4c55ec91dacb1e3145f9dbe022503d07e4b909ee1a5d8R459)

kacperkapusciak (approved):
Sorry, @denisk20! You're right 😓 I should've checked the whole context.

PR #1120, opened 2022-02-16
fix: prevent-going-back snack crash

kacperkapusciak:
## Description

An incorrect component import in a `prevent-going-back` Snack leads to a crash.

Changing `TextEditText` ➡️  `TextInput` fixes the problem.


eriveltonelias (approved):
Thanks!

PR #1135, opened 2022-03-18
docs: Use message.data instead of message.notification for Firebase Cloud Messaging

bangseongbeom:
- Use data instead of notification

    [Notification](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#notification) does not support `url` property.

- Use optional chaining operators more precisely

- Replace Linking.removeEventListener

    `Linking.removeEventListener` is deprecated.

    See: https://reactnative.dev/docs/next/linking#removeeventlistener

- Add @react-native-firebase/messaging imports


satya164 (approved):
Thanks for the PR

PR #1136, opened 2022-03-20
docs: update link text

mdneyazahmad:
This pr updates the link text `NavigationActions` to `CommonActions` in `TabActions` to be consistent across other `Action` page. `StackActions` and `DrawerActions` has the same content with `CommonActions` as link text.


kpose:
Hi @mdneyazahmad , thanks for this. However, as this bug is also present in version 5.x, do you mind adding the fix for that to this PR? I will be happy to take a look

kpose (approved):

PR #1139, opened 2022-03-26
Change migration guide url and text

NaseelNiyas:
## Change the text and url of the migration guide on the Splash screen. 

Current: upgrade from **v4** to **v5**
Changes: upgrade from **v5** to **v6**


satya164:
Thanks for the PR! However we've updated this link to list migration guides for several versions instead, so I'm closing it

PR #1141, opened 2022-03-30
interpolate to interpolateNode

LimeyMoon:
React native reanimated newer version changed the name interpolate to interpolateNode.
interpolate function doesn't work now.
Need to change code for the snack too so that it can work.


satya164:
Thanks for the PR!

PR #1145, opened 2022-04-19
Fix types organization example

xMartin:
This works for me and it's also consistent with the initial example in "Combining navigation props".


satya164 (changes requested):
This is incorrect. As defined earlier in the snippet, `RootStackScreenProps` takes a screen name, not a param list.

> consistent with the initial example

They are different types

xMartin:
You are right. I confused them. Thanks for reviewing!

PR #1155, opened 2022-06-20
fix: fix SafeArea imports

EduVencovsky:
Looking at other places in the docs, I realized that this code snippet have a wrong import


satya164 (approved):
Thank you

PR #1156, opened 2022-06-21
Update deep-linking.md

amiinous:
According to official docs (https://developer.android.com/training/app-links/verify-site-associations), the scheme and host should be defined in separate `data` tags. Initially I followed the current doc, but after doing some search I found out that combining the `scheme` and `host` into one `data` broke the universal links in Android.


satya164:
Thank you

satya164 (approved):

satya164 (changes requested):
I think there needs to be a different intent filter for this case, otherwise it'll need to match `mychat://www.example.com` which won't be correct

amiinous:
@satya164 You're right...I separated the two deep-linking schemes.

satya164 (approved):
Thank you

PR #1165, opened 2022-07-16
update react-native-gesture-handler install doc

ramyareye:


kacperkapusciak:
Hi @ramyareye :wave:

I don't this change is necessary as we wrap stack & drawer with `GestureHandlerRootView` internally in react-navigation eg. [here](https://github.com/react-navigation/react-navigation/blob/f54cafbb3333763ec86e4b2cab46cbb2ae99b627/packages/stack/src/views/Stack/StackView.tsx#L429). 

Have you had problems with installing `react-native-gesture-handler` in `react-navigation` and wrapping your app with  `GestureHandlerRootView` helped with?

Cheers

ramyareye:
> Hi @ramyareye 👋
> 
> I don't this change is necessary as we wrap stack & drawer with `GestureHandlerRootView` internally in react-navigation eg. [here](https://github.com/react-navigation/react-navigation/blob/f54cafbb3333763ec86e4b2cab46cbb2ae99b627/packages/stack/src/views/Stack/StackView.tsx#L429).
> 
> Have you had problems with installing `react-native-gesture-handler` in `react-navigation` and wrapping your app with `GestureHandlerRootView` helped with?
> 
> Cheers

there is a step to finalize `react-native-gesture-handler` installation in [here](https://reactnavigation.org/docs/drawer-navigator/#installation) and [here](https://reactnavigation.org/docs/stack-navigator/#installation) which is outdated way to setup, either remove them or fix? 
Anyway I didn't read you have this `GestureHandlerRootView` under the hood anywhere :D

satya164:
Those steps are needed if you use gesture handler API in your project. For React Navigation, it already takes care of adding the root view whenever necessary and the consumer doesn't need to add them manually.

PR #1167, opened 2022-08-06
Use existing icon in example for Tab Navigation

r3nya:
# Info

There is no `ios-list-box` icon in `react-native-vector-icons/Ionicons` icon set. You can check it using https://oblador.github.io/react-native-vector-icons/

That's why using `ios-list` and `ios-list-outline` much more better.


satya164 (approved):
Thank you

PR #1168, opened 2022-08-20
docs: Added a guide to install react-native-reanimated in drawer-navigator.md

ajatdarojat45:
# Info

i just added a guide to install react-native-reanimated in drawer-navigator.md


satya164:
Thanks for the PR. The detailed installation instructions are already mentioned in Reanimated's docs so I don't see the need to duplicate it and potentially get outdated. If you think the link to reanimated docs should be emphasized then ideas welcome.

PR #1170, opened 2022-08-24
Update getting-started.md

Gregoirevda:


satya164:
This is not correct. It's explicitly set to `null` as per React Native Screens' docs: https://github.com/software-mansion/react-native-screens#android

PR #1175, opened 2022-09-12
changed file extension from .tsx to .ts for types file

Adnan-Bacic:
.tsx files should only be used for react components containing jsx, while .ts files should be for functions, types etc.

in this case, the `types` file will only contain types for the navigation.


satya164:
`.tsx` and `.ts` files have some differences in the supported syntax (e.g. the legacy type assertion syntax works in `.ts` files). We use `.tsx` for everything to avoid any syntactic ambiguity and for simplicity so we don't need to think about what extension to use for each file.

Most people use `.tsx` for React components only, but there's no specific reason that it should only be used for React components.

PR #1180, opened 2022-10-10
docs: fixes component name in Navigating to a screen in a nested navi…

mikemikhaylov:
Changed component name from `Settings` to `Feed` since screen name is `Feed` and the component referred as `Feed` in the text below.


satya164 (approved):
Thank you

PR #1182, opened 2022-10-17
Make drawer installation steps more clear

ddeath:
With this change we will highlight that all 3 steps are required. In old version it looked to me that 3rd step could be only for bare react native project


satya164 (approved):
Thank you

PR #1183, opened 2022-10-20
docs: update Screen Tracking example to remove expo-firebase-analytics

amandeepmittal:
## Why and how

[expo-firebase-analytics](https://docs.expo.dev/versions/latest/sdk/firebase-analytics/) is getting [deprecated](https://github.com/expo/expo/pull/19398). With SDK 47 going to be released soon, we want developers to migrate from it and use any other mobile analytics SDK.

This PR 
- updates remove references to using `expo-firebase-analytics` from the example in Screen Tracking
- adds a placeholder to indicate at which line the analytics function should be added when using the example
- update the verbiage in context to example
- updates version `6.x` docs and changes backported for `5.x`.

### DIFF
--- a/versioned_docs/version-5.x/screen-tracking.md
+++ b/versioned_docs/version-5.x/screen-tracking.md
@@ -32,6 +32,11 @@ export default () => {
       onStateChange={async () => {
         const previousRouteName = routeNameRef.current;
         const currentRouteName = navigationRef.current.getCurrentRoute().name;
+        const tracker = {
+          trackScreenView: () => {
+            // Your implementation of analytics goes here!
+          };
+        }
 
         if (previousRouteName !== currentRouteName) {
           // Replace the line below to add the tracker from a mobile analytics SDK
--- a/versioned_docs/version-5.x/screen-tracking.md
+++ b/versioned_docs/version-5.x/screen-tracking.md
@@ -32,15 +32,13 @@ export default () => {
       onStateChange={async () => {
         const previousRouteName = routeNameRef.current;
         const currentRouteName = navigationRef.current.getCurrentRoute().name;
-        const tracker = {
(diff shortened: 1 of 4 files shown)


brentvatne (approved):

satya164 (changes requested):
Thanks for the PR. Can you update the sample to do a `console.log` instead of removing it?

Also, I'd remove the `tracker` object and create a `trackScreenView` function directly for the examples

amandeepmittal:
Thanks for the feedback @satya164!

I've removed the `tracker` object reference and converted the `trackScreenView` to a function for the example snippets.

The code samples to Expo Snack are referenceed back. I found references to `expo-firebase-analytics` (commented out) in the samples, so I removed it (as not to create any confusion about the package's availability).

Also, in the sample files, there is already an `alert()` that displays the current route name as a placeholder, so didn't change that to `console.log`. I think it conveys a similar behavior for sample examples.

versioned_docs/version-5.x/screen-tracking.md
```diff
@@ -13,12 +13,9 @@ To get notified of state changes, we can use the `onStateChange` prop on `Naviga
 
 ## Example
 
-This example shows how to do screen tracking and send to Firebase Analytics using [expo-firebase-analytics](https://docs.expo.io/versions/latest/sdk/firebase-analytics/). The approach can be adapted to any other analytics SDK.
-
- <samp id="screen-tracking-for-analytics" />
```
satya164, line 18: Can you add this back here too?

amandeepmittal: I missed that. Sorry, my bad 🙏 

Added it back.

satya164 (approved):
Thanks!

PR #1187, opened 2022-11-10
More clarification with location of override

evan1715:
Provided more clarification to the location of where the override location should be. This will help users who are experiencing crashes still because they have placed it in the MainAcitivityDelegate and not the parent function.

[rn screens users - #1481](https://github.com/software-mansion/react-native-screens/issues/1481)


kacperkapusciak:
Thanks for pointing that out!

Fixed in 72ea67a7032ede731294fc70fb3d448989e18d61

PR #1189, opened 2022-11-23
docs: add tintColor option for headerSearchBarOptions

luoxuhai:
Added missing `tintColor` option for `headerSearchBarOptions`.

`tintColor`: https://github.com/software-mansion/react-native-screens/blob/main/src/types.tsx#L472

**Example**
`tintColor: "#FF0000"`:

image


kacperkapusciak (approved):
Thank you! 🙏

PR #1196, opened 2022-12-23
docs: better function description

pvinis:
Better way to describe this function.


kacperkapusciak:
Hi @pvinis 👋 
Thanks for pointing that out! It should've been `createXNavigator` instead of `createNavigatorX`.

We're using `createXNavigator` naming [since 2018](https://reactnavigation.org/blog/2018/04/06/react-navigation-2.0-rc#xnavigator-is-now-named-createxnavigator) and we aren't keen on changing that to an * symbol 😅 

Cheers

PR #1201, opened 2023-01-09
Fix incorrect wrapping causing 'jumpy' headers

scarlac:
Fixes issues like #9557 where users aren't explicitly using `react-native-safe-area-context` (but it's a requirement for React Navigation). The onboarding/getting-started docs should exemplify this requirement so we don't get reports and glitchy/jumpy outcomes for apps using the library

### DIFF
--- a/versioned_docs/version-6.x/getting-started.md
+++ b/versioned_docs/version-6.x/getting-started.md
@@ -94,10 +94,13 @@ Now, we need to wrap the whole app in `NavigationContainer`. Usually you'd do th
 ```js
 import * as React from 'react';
 import { NavigationContainer } from '@react-navigation/native';
+import { SafeAreaProvider } from 'react-native-safe-area-context';
 
 export default function App() {
   return (
-    <NavigationContainer>{/* Rest of your app code */}</NavigationContainer>
+    <SafeAreaProvider>
+      <NavigationContainer>{/* Rest of your app code */}</NavigationContainer>
+    </SafeAreaProvider>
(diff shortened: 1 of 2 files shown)


satya164:
This isn't a requirement. If you have jumpy behavior then it's either due to a bug in `react-native-safe-area-context` giving inconsistent values between `initialWindowMetrics` and `SafeAreaProvider` or your status bar is changing between translucent and opaque. If you don't have the later, please open a bug report in safe area context repo with a repro.

scarlac:
I never claimed this was a bug in the library? This is a clear transitive requirement from the safe area library.

I have tested this locally and there was no issues with header coloration nor sizing.

It's just that the examples are showing code that will lead to issues if copy pasted as-is which is misleading and causes these people to report issues. 

If you look at my comment I'm that thread you can see I'm referencing a comment from another user who's comment has been "upvoted".

Please reconsider this doc change. It will prevent further confusion for others.

satya164:
> I never claimed this was a bug in the library?

I didn't say you claimed it's a bug. I said that it might be a bug if you're facing jumpiness.

> This is a clear transitive requirement from the safe area library.

React Navigation already uses the provider - otherwise, it'd throw an error and crash. There's no requirement to wrap your app in `SafeAreaProvider` and no one should need to.

It's very intentional that this isn't mentioned in the doc, this is intended to work without manual wrapping by the user. This isn't a documentation issue - it is a bug and needs to be fixed.

PR #1202, opened 2023-01-15
Add cleanup function to 'prevent going back' examples and docs

charliedieter:
I noticed after copying the example that without a cleanup function, multiple instances of this event are sometimes created. I'm relatively new to RN, so let me know if this is incorrect :). 

- [x]  Make sure to add your changes to versioned docs


satya164:
Thanks for the PR, however the current examples already return the cleanup function.

```js
const foo = () => doSomething();
```

is same as

```js
const foo = () => {
  const bar = doSomething();

  return bar;
}
```

charliedieter:
😅 apologies for the spam - I completely missed the implicit return.

PR #1229, opened 2023-04-09
Add note how to use KeyboardAvoidingView with tab and stack navigator

tomwaitforitmy:
Details are described here: https://github.com/react-navigation/react-navigation/issues/11244


stuckj (approved):
Using this advice fixed the keyboard for our app. Trying to track this down was a nightmare.

tomwaitforitmy:
I was having doubts that I missed anything crucial, because nobody showed any interest here. But I definitely had a minimal repro and fixing stuff with KeyboardAvoidingView is always pain 😅. Thanks for sharing! I'm glad that it helped you.

PR #1236, opened 2023-04-21
Update material-top-tab-navigator.md

Smartech-git:
Adding "overScrollMode" prob to the Tab-Navigator removes and setting to "false" removes the overScrollMode animation in Andriod. I think Should be added in the docs.


okwasniewski:
Thanks for creating this PR. However, this prop is not added in typescript types so if you want to add this behavior to docs then you would also need to update the types.

satya164 (approved):
Thanks for the PR

PR #1239, opened 2023-05-03
Update stack-actions.js

DomiR:
Fix stack action examples for future version. The button labels on the example were swapped.


stanleyoos (approved):
Good catch! 👀

PR #1250, opened 2023-06-08
Fix typo in stack-navigator and native-stack-navigator for `replace` example

josepdecid:
The code examples for `replace` in both `stack-navigator` and `native-stack-navigator` were using the `push` method instead of `replace`.


lord-Sid (approved):

kacperkapusciak (approved):
Thank you! 🙇

PR #1254, opened 2023-06-15
Update stack-navigator.md

kazuma0129:
It seems to show an example of a code that is different from what is described.


kacperkapusciak (approved):
Thank you! 🙇

PR #1263, opened 2023-07-12
Update hiding-tabbar-in-screens with style option.md

FrederickEngelhardt:
- adds a second option using react-native styles to hide the bottom bar.
- The current example (moving the stack above the tabs) shows ideal use cases while the second option works when those ideal use cases are not an option.

# READ ME PLEASE

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-6.x/hiding-tabbar-in-screens.md
+++ b/versioned_docs/version-6.x/hiding-tabbar-in-screens.md
@@ -59,3 +59,55 @@ function App() {
 ```
 
 After re-organizing the navigation structure, now if we navigate to the `Profile` or `Settings` screens, the tab bar won't be visible over the screen anymore.
+
+### Alternative Options - Hiding the bar via stylea
+
+Use cases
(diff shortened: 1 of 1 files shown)


versioned_docs/version-6.x/hiding-tabbar-in-screens.md
```diff
@@ -59,3 +59,55 @@ function App() {
 ```
 
 After re-organizing the navigation structure, now if we navigate to the `Profile` or `Settings` screens, the tab bar won't be visible over the screen anymore.
+
+### Alternative Options - Hiding the bar via stylea
```
FrederickEngelhardt, line 63:
```suggestion
### Alternative Options - Hiding the bar via styles
```

FrederickEngelhardt:
@satya164 Are you able to merge this? This doc update gives an example how to hide bottom bar vs the previous example was a tutorial how to **not** hide it by refactoring a screen approach.

PR #1264, opened 2023-07-14
Example consistency fix

peelle:
Thank's for taking the time to review. This change matches the example code in other sections like v5 linkto, and the example of getPathFromState right underneath the modified code. As an infrequent JS coder the inconsistency caused me some confusion. Hopefully this change will help others in my position.


satya164 (approved):
Thanks for the PR

PR #1266, opened 2023-07-19
Update deep-linking.md - Fix Declare Association between website and …

BismarkCodes:
…intent-filters external documentation url

The current link redirects to https://support.google.com/android/search?q=open+by+default#web-assoc which does not provide the intended document. I changed it to the actual url which is 'https://developer.android.com/training/app-links/verify-android-applinks#web-assoc'

# READ ME PLEASE

Thanks for opening a PR!

### DIFF
--- a/versioned_docs/version-6.x/deep-linking.md
+++ b/versioned_docs/version-6.x/deep-linking.md
@@ -185,7 +185,7 @@ After adding them, it should look like this:
 </activity>
 ```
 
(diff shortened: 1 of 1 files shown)


mattpmoreno:
@BismarkCodes is your update still true? When I look up instructions on how to add association, Google tells me to use Google Play Store -> App -> Setup -> Advanced Settings -> App Indexing -> Verified Websites. Not pushing back on your PR, just trying to figure out what the right way to do this is... Thanks!

PR #1267, opened 2023-07-25
Update navigation-prop.md

adamsythe:
It's neater with null chaining

# READ ME PLEASE

Thanks for opening a PR!


satya164:
Thanks for the PR. However 3.x is quite old and I'm not sure if optional chaining was available in React Native back then. So I'd prefer to preserve the documentation to how it was, nobody should be using it with latest React Native anyway.

So I'm closing this.

Issue #1270, opened 2023-08-15
Improve React-Navigation SVGs


RiteshK-611:
### Problem
All react-navigation SVGs are not optimized properly. And they have big file size of around 33 KB which is because these SVGs has many control points.
**SVG Files**
https://github.com/react-navigation/react-navigation.github.io/blob/main/static/img/spiro.svg
https://github.com/react-navigation/react-navigation.github.io/blob/main/static/img/spiro_header.svg
https://github.com/react-navigation/react-navigation.github.io/blob/main/static/img/spiro_white.svg

### Solution
This problem can be solve if we can reprocess the icon with curves rather than one-by-one control points.

panditakshay402:
@RiteshK-611 can you assign this issue!

PR #1271, opened 2023-08-16
Update/Improve React-Navigation SVGs

RiteshK-611:
Issue Close: #1270

This change resolves issue #1270 by updating following **React-Navigation** SVGs 
https://github.com/react-navigation/react-navigation.github.io/blob/main/static/img/spiro.svg
https://github.com/react-navigation/react-navigation.github.io/blob/main/static/img/spiro_header.svg
https://github.com/react-navigation/react-navigation.github.io/blob/main/static/img/spiro_white.svg

This change results in 90% decrease in file size (33 --> 3.3)


RiteshK-611:
Hi @satya164 
It would be great if you review this PR? 🙂

PR #1293, opened 2023-12-19
Rework fundamentals to accomodate static API

satya164:


versioned_docs/version-7.x/getting-started.md
```diff
@@ -114,9 +117,25 @@ When you use a navigator (such as stack navigator), you'll need to follow the in
...
+Once you've installed and configured the dependencies, you can move on to setting up your project to use React Navigation.
+
+When using React Navigation, you configure [**navigators**](glossary-of-terms.md#navigator) in your app. Navigators handle the transition between screens in your app and provide UI such as header, tab bar etc.
+
+There are 2 primary ways to configure the navigators:
+
+### Static configuration
 
-Now, we need to wrap the whole app in `NavigationContainer`. Usually you'd do this in your entry file, such as `index.js` or `App.js`:
+The static configuration API has reduced boilerplate and simplifies things such as TypeScript types and deep linking. If you're starting a new project or are new to React Navigation, this is the **recommended way** to set up your app.
```
osdnk, line 130: I'm somehow unconvinced about stating that this is "the recommended way". I think we can just present both with their tradeoffs?

versioned_docs/version-7.x/use-route.md
```diff
@@ -24,7 +24,7 @@ function MyText() {
 }
 ```
 
-See the documentation for the [`route` prop](route-prop.md) for more info.
+See the documentation for the [`route` prop](route-object.md) for more info.
```
osdnk, line 27:
```suggestion
See the documentation for the [`route` object](route-object.md) for more info.
```

versioned_docs/version-7.x/typescript.md
```diff
@@ -231,7 +231,7 @@ const navigation = useNavigation<ProfileScreenNavigationProp>();
 :::danger
 
 Annotating `useRoute` isn't type-safe because the type parameter cannot be statically verified.
-Prefer using the [`route` prop](route-prop.md) instead when possible. Use `useRoute` for generic code that doesn't need specific route type.
+Prefer using the [`route` prop](route-object.md) instead when possible. Use `useRoute` for generic code that doesn't need specific route type.
```
osdnk, line 234:
```suggestion
Prefer using the [`route` object](route-object.md) instead when possible. Use `useRoute` for generic code that doesn't need specific route type.
```

versioned_docs/version-7.x/static-configuration.md
```diff
@@ -1,7 +1,7 @@
 ---
 id: static-configuration
-title: Getting started with static API
-sidebar_label: Getting started
+title: Overview of static API
+sidebar_label: Overview
 ---
 
 React Navigation primarily uses a dynamic component based API. This provides a lot of flexibility, however there are a few drawbacks:
```
osdnk, line 7: This narrative is misleading. As we first said, the static API is recommended, but this one is "primarily used". It's a bit confusing.

versioned_docs/version-7.x/headers.md
```diff
@@ -114,12 +193,40 @@ There are a couple of things to notice here:
...
+It is common to want to configure the header in a similar way across many screens. For example, your company brand color might be red and so you want the header background color to be red and tint color to be white. Conveniently, these are the colors we're using in our running example, and you'll notice that when you navigate to the `DetailsScreen` the colors go back to the defaults. Wouldn't it be awful if we had to copy the `options` header style properties from `Home` to `Details`, and for every single screen we use in our app? Thankfully, we do not. We can instead move the configuration up to the native stack navigator under `screenOptions`:
```
osdnk, line 196:
```suggestion
It is common to want to configure the header in a similar way across many screens. For example, your company brand color might be red and so you want the header background color to be red and the tint color to be white. Conveniently, these are the colors we're using in our running example, and you'll notice that when you navigate to the `DetailsScreen` the colors go back to the defaults. Wouldn't it be awful if we had to copy the `options` header style properties from `Home` to `Details`, and for every single screen we use in our app? Thankfully, we do not. We can instead move the configuration up to the native stack navigator under `screenOptions`:
```

I think it's a concrete color.

versioned_docs/version-7.x/getting-started.md
osdnk, line 130:
```suggestion
The static configuration API has reduced boilerplate and simplifies things such as TypeScript types and deep linking. It might be considered a good way to quickly start the development of a new app, not putting too much energy into configuring the navigation structure. Although it may not provide a sufficient level of flexibility, that should be more than enough for the majority of the applications. 
```

versioned_docs/version-7.x/glossary-of-terms.md
```diff
@@ -55,41 +55,32 @@ const StackNavigator = (
...
 - `navigate`, `goBack`, etc are available to dispatch actions in a convenient way
 
-Navigators can also accept a navigation prop, which they should get from the parent navigator, if there is one.
+This object can be accessed with the [`useNavigation`](use-navigation.md) hook. It's also passed as a prop to screens defined with the dynamic API.
 
-For more details, see the ["Navigation prop document"](navigation-prop.md).
+For more details, see the ["Navigation object docs"](navigation-object.md).
 
-The ["Route prop reference"](route-prop.md) section goes into more detail on this, describes workarounds, and provides more information on other properties available on `route` prop.
+The ["Route object reference"](route-object.md) section goes into more detail on this, describes workarounds, and provides more information on other properties available on `route` prop.
```
osdnk, line 69:
```suggestion
The ["Route object reference"](route-object.md) section goes into more detail on this, describes workarounds, and provides more information on other properties available on `route` object.
```

versioned_docs/version-7.x/navigation-events.md
```diff
@@ -93,7 +93,7 @@ class Profile extends React.Component {
 }
 ```
 
-One thing to keep in mind is that you can only listen to events from the immediate navigator with `addListener`. For example, if you try to add a listener in a screen that's inside a stack that's nested in a tab, it won't get the `tabPress` event. If you need to listen to an event from a parent navigator, you may use [`navigation.getParent`](navigation-prop.md#getparent) to get a reference to parent navigator's navigation prop and add a listener.
+One thing to keep in mind is that you can only listen to events from the immediate navigator with `addListener`. For example, if you try to add a listener in a screen that's inside a stack that's nested in a tab, it won't get the `tabPress` event. If you need to listen to an event from a parent navigator, you may use [`navigation.getParent`](navigation-object.md#getparent) to get a reference to parent screen's navigation object and add a listener.
```
osdnk, line 96:
```suggestion
One thing to keep in mind is that you can only listen to events from the immediate navigator with `addListener`. For example, if you try to add a listener in a screen that's inside a stack that's nested in a tab, it won't get the `tabPress` event. If you need to listen to an event from a parent navigator, you may use [`navigation.getParent`](navigation-object.md#getparent) to get a reference to the parent screen's navigation object and add a listener.
```

versioned_docs/version-7.x/static-configuration.md
```diff
@@ -1,7 +1,7 @@
 ---
 id: static-configuration
-title: Getting started with static API
-sidebar_label: Getting started
+title: Overview of static API
```
osdnk, line 3: I think it should come with with article.

versioned_docs/version-7.x/navigation-events.md
osdnk, line 96:
```suggestion
One thing to keep in mind is that you can only listen to events from the immediate navigator with `addListener`. For example, if you try to add a listener in a screen that's inside a stack that's nested in a tab, it won't get the `tabPress` event. If you need to listen to an event from a parent navigator, you may use [`navigation.getParent`](navigation-object.md#getparent) to get a reference to the parent screen's navigation object and add a listener.
```

versioned_docs/version-7.x/getting-started.md
```diff
@@ -114,9 +117,25 @@ When you use a navigator (such as stack navigator), you'll need to follow the in
...
+
+Continue to ["Hello React Navigation"](hello-react-navigation.md?config=static) to start writing some code with the static API.
+
+### Dynamic configuration
+
+The dynamic configuration allows for more flexibility but requires more boilerplate and configuration.
```
osdnk, line 136:
```suggestion
The dynamic configuration allows for more flexibility but requires more boilerplate and configuration. 
Use that option if you would like to manage your routes dynamically and need the flexibility of dealing with the React components directly. 

Also, this used to be a default way of defining the navigation structure before the 7.x version. If you are updating from the older version, we do not suggest migration, unless the level of boilerplate was causing difficulties. 
```

versioned_docs/version-7.x/static-configuration.md
satya164: Gonna delete this page since now fundamentals section covers all of these.

PR #1310, opened 2024-02-22
Migrate 'Type checking with TypeScript' examples to static API in v7

patrycjakalinska:


kacperkapusciak (changes requested):
I'm not sure whether this PR is true. In Static API Typescript is mostly inferred when properly set-up. This was explained in https://reactnavigation.org/docs/7.x/static-typescript

So I don't think we need this in this form?

satya164:
Yeah, I think we should move the whole page content of https://reactnavigation.org/docs/7.x/static-typescript and basically the tabs will contain the full content of the page.

We should do similar for https://reactnavigation.org/docs/7.x/static-authentication as well.

Then https://reactnavigation.org/docs/7.x/static-combine-with-dynamic can be moved to under "Guides".

The remaining one https://reactnavigation.org/docs/7.x/static-api-reference can be then moved to under the "API references" section and renamed to "Static API reference"

These pages were written before we decided with the tab approach so that's why they are on different pages right now.

kacperkapusciak:
Superseded by 6765a7cab51103f8e69441d97426dad96d6d2d3f

PR #1315, opened 2024-02-23
Migrate 'Screen tracking for analytics' examples to static API

patrycjakalinska:
Needs adjustment for when the ref is fixed.


kacperkapusciak (approved):
🚀

PR #1319, opened 2024-02-24
Migrate 'Groups' examples to static API in v7

patrycjakalinska:


kacperkapusciak (approved):
It's weird to me API-wise that both `screens` and `groups` have to be defined but the examples are correct

PR #1320, opened 2024-02-24
Migrate 'Navigation Container' examples to static API in v7

patrycjakalinska:


kacperkapusciak (approved):
All the examples look good but there's no longer a `NavigationContainer` on Static API anymore so the page content is a bit off

We can update it later on tho

PR #1323, opened 2024-02-26
Migrate 'Screen' examples to static API in v7

patrycjakalinska:
Add static examples - exluding `children` example


kacperkapusciak (approved):
Same as with `NavigationContainer` there's no `Screen` component anymore in Static API in https://github.com/react-navigation/react-navigation.github.io/pull/1320

All examples are correct tho

PR #1341, opened 2024-05-08
Fix screen-options.md typo

crisanmm:
this fixes a possible typo: `to add specify` → `to specify`

I only found this typo in `version-6.x` and `version-7.x` so I updated it only there


nmiz1987 (approved):
the 'add' should not be there.

nmiz1987 (approved):
the 'add' should not be there.

stanleyoos (approved):

PR #1343, opened 2024-05-24
Update upgrading-from-6.x.md (minor typo)

robwalkerco:
When reading the V7 upgrade guide, what appears to be an accidently added word caused me some confusion.


stanleyoos (approved):
Nice catch! Thank you

PR #1352, opened 2024-08-01
chore: Native Stack doc's - headerBackButtonMenuEnabled

stanleyoos:
React Navigation, Native Stack, v7.x

- added screenshots for: `headerBackTitle`, `headerBackTitleStyle`, `headerBackButtonMenuEnabled`


static/assets/7.x/native-stack/animation/native-stack-animation-default.mp4
kacperkapusciak, line 1: I don't think it's necessary to split these files into separate folder, even worse having a separate folder for a singular image eg. `headerBackTitle/headerBackTitle.jpeg`. A better approach is to have it all in a single `native-stack` folder and using names for files that group them well.

versioned_docs/version-7.x/native-stack-navigator.md
```diff
@@ -193,6 +197,31 @@ Style object for header back title. Supported properties:
...
+          headerBackTitle: 'Back title example',
+          headerBackTitleStyle: {
+            fontSize: 14,
+            fontFamily: 'Georgia',
+          },
+        }}
+      />
+    </Stack.Navigator>
+  );
+}
```
kacperkapusciak, line 222: If you want to put a full example here you'd also have to add Static API (your example is in Dynamic API) see https://reactnavigation.org/docs/7.x/native-stack-navigator#usage for reference

Here. maybe a simpler example would also work just fine?

```suggestion
headerBackTitleStyle: {
  fontSize: 14,
  fontFamily: 'Georgia',
},
```

kacperkapusciak (approved):
perfect!

PR #1362, opened 2024-08-07
docs: Updated references to React documentation

stanleyoos:


versioned_docs/version-6.x/getting-started.md
```diff
@@ -12,7 +12,7 @@ If you're already familiar with JavaScript, React and React Native, then you'll
 
 Here are some resources to help you out:
 
-1. [React Native Express](https://www.reactnative.express) (Sections 1 to 4)
+1. [React Native](https://reactnative.dev/docs/getting-started) (Sections 1 to 4)
```
tboba, line 15: Why are we changing this reference?

tboba:
Also, I see in some places in code there are still references to reactjs.org. Should we also change them?

stanleyoos:
@tboba I updated only v6 and v7 intentionally, since v5 is ~4 years old, do you think we should update all links?

tboba:
No no, I mean that in `6.x` and `7.x` versions there are still references to reactjs.org, like in `function-after-focusing-screen.md` article, there are references to [intro for hooks](https://reactjs.org/docs/hooks-intro.html) guide. If you're using VSCode, you can cmd-shift-f on the project and look for reactjs.org - this way you could check if any articles from `versioned_docs/version-6.x` or `versioned_docs/version-7.x` still contain any references to reactjs.org 😄

versioned_docs/version-6.x/function-after-focusing-screen.md
```diff
@@ -47,7 +47,7 @@ In most cases, it's recommended to use the `useFocusEffect` hook instead of addi
 
 ## Triggering an action with the `useFocusEffect` hook
 
-React Navigation provides a [hook](https://reactjs.org/docs/hooks-intro.html) that runs an effect when the screen comes into focus and cleans it up when it goes out of focus. This is useful for cases such as adding event listeners, for fetching data with an API call when a screen becomes focused, or any other action that needs to happen once the screen comes into view.
+React Navigation provides a [hook](use-focus-effect.md) that runs an effect when the screen comes into focus and cleans it up when it goes out of focus. This is useful for cases such as adding event listeners, for fetching data with an API call when a screen becomes focused, or any other action that needs to happen once the screen comes into view.
```
tboba, line 50: I believe that's incorrect reference right now and we rather want to redirect user to something like [this](https://react.dev/reference/react/hooks) 🤔

versioned_docs/version-6.x/function-after-focusing-screen.md
```diff
@@ -75,7 +75,7 @@ See the [`useFocusEffect`](https://reactnavigation.org/docs/use-focus-effect/) d
 
 ## Re-rendering screen with the `useIsFocused` hook
 
-React Navigation provides a [hook](https://reactjs.org/docs/hooks-intro.html) that returns a boolean indicating whether the screen is focused or not.
+React Navigation provides a [hook](use-is-focused.md) that returns a boolean indicating whether the screen is focused or not.
```
tboba, line 78: Same here aswell

tboba (changes requested):

versioned_docs/version-6.x/getting-started.md
tboba: Yeah, I see that last commit in React Native Express was about 3 years ago so we can proceed with that, but we should remove `(Sections 1 to 4)` then 🤔

versioned_docs/version-6.x/function-after-focusing-screen.md
stanleyoos: The official React documentation does not have any details about [`useFocusEffect`](https://reactnavigation.org/docs/7.x/use-focus-effect/) and [`useIsFocused`](https://reactnavigation.org/docs/7.x/use-is-focused/), unlike React Navigation, where we have corresponding pages for that.
Should we still change these references?

tboba: Oh, I see what you did here 😄 I think it's better to redirect to useFocusEffect than to intro to the hooks, great change though

tboba (approved):
🎉

PR #1374, opened 2024-10-07
Update testing section

AnCichocka:
update testing with `jest` section:
- add mocking `react-native-screens`
- add example tests


versioned_docs/version-7.x/testing.md
```diff
@@ -54,53 +48,257 @@ Then we need to use this setup file in our jest config. You can add it under `se
...
+
+  // All exports in react-native-screens are getters
+  // We cannot use spread for cloning as it will call the getters
+  // So we need to clone it with Object.create
+  screens = Object.create(
+    Object.getPrototypeOf(screens),
+    Object.getOwnPropertyDescriptors(screens)
+  );
+
+  return screens;
```
kkafar, line 67: We are missing here example of actual mock of some component from `react-native-screens`. Try to mock `Screen` as a `View` for example.

kkafar (commented):
Hey, I think the examples look nice, we might want to get some insights from @satya164 here also

versioned_docs/version-7.x/testing.md
```diff
@@ -54,53 +48,670 @@ Then we need to use this setup file in our jest config. You can add it under `se
...
+  expect(renderCounter).toBe(1);
+});
+```
+
+</TabItem>
+</Tabs>
+
+For writing tests that include times functions we will have to use [Fake Timers](https://jestjs.io/docs/timer-mocks). They will replace times function implementation to use time from the fake clock.
+
+Let's add another button to the Profile screen, which uses `setTimeout`:
```
AnCichocka, line 591: If we decide on leaving preload test as it is with navigator definition and test in one file, I will adjust this section text, because it's based on previous navigator

satya164 (changes requested):
Thanks for the PR!

I think the current examples showcasing different navigation methods aren't the best examples to use here. The important thing to note that this documentation is for users of the library. These tests showing how different navigation methods work isn't what the user will need. Our docs should be more focused on what users need.

Similarly, the fake timers examples are generic fake timer example code. The idea is to demonstrate why it maybe necessary in some cases when writing tests for navigation code (e.g. it's necessary in case of JS stack navigator because navigating triggers an animation, so with fake timers, we can instantly skip the animation for testing).

Here are some examples:

1. Navigating to a screen and asserting that the new screen content is rendered.
   - By pressing a button
   - By pressing tabs on a tab bar
   - By pressing drawer item
2. Testing app's functionality when using various hooks from React Navigation
   - Check if correct content is rendered when screen is focused and unfocused (using `useIsFocused` hook)
   - Check if data is fetched/a side effect is performed when screen comes to focus (using `useFocusEffect` hook)
3. Testing app's functionality with code in various events (e.g. `tabPress`)

You can also think of more cases like this - specifically, what kind of tests you may write when building an app that involves navigation.

We should structure the docs in this way:

- Problem statement - example of a real-world case that the user may want to test
- Example code
- Explanation for the example code
- Explanation for any special code we may have done, e.g. we use fake timers to skip animations in JS stack navigators

versioned_docs/version-7.x/testing.md
```diff
@@ -54,59 +48,855 @@ Then we need to use this setup file in our jest config. You can add it under `se
...
+import { createStaticNavigation } from '@react-navigation/native';
+import { fireEvent, render, screen } from '@testing-library/react-native';
+
+import { StackNavigator } from './StackNavigator';
+
+test('navigates to settings by "Go to Settings" button press', () => {
+  const StackNavigation = createStaticNavigation(StackNavigator);
+  render(<StackNavigation />);
+
+  fireEvent.press(screen.queryByText('Go to Settings'));
```
przemek-gawron, line 181: Could you use `getBy…` instead of `queryBy…` when we expect that an element will be present on the screen, and use `queryBy…` only when we expect that an element won’t be available?
- This way, it will be clearer which elements we should expect to be visible or not.
- In case of an error, `getBy…` will fail earlier than `queryBy….`

PR #1378, opened 2024-10-25
typo

pvinis:
I'm guessing this is what we mean here, and not some hook.


satya164:
As mentioned in the paragraph below, it takes a hook.

PR #1384, opened 2024-11-15
Replace removed `labelVisible` with `displayMode`.

KirillTregubov:
I was migrating to React Navigation v7 and I noticed that the docs were a bit outdated. I referred to [this elements changelog](https://github.com/react-navigation/react-navigation/blob/main/packages/elements/CHANGELOG.md#200-rc19-2024-08-05) and [this Native Stack Navigator docs page](https://reactnavigation.org/docs/native-stack-navigator/#headerbackbuttondisplaymode) when making my changes.

I updated the elements docs page itself, and the migration guide to help others with their migrations. Feel free to move it to a different category, as I created a new one for it `"Changes to elements"`.

Please verify my formatting before accepting. I noticed you usually put the default value after the possible values, but I wasn't able to figure out how to do that within the list.

Thanks for the awesome library, I hope I was able to help fix the issue!


satya164 (approved):
Thanks for the PR!

PR #1387, opened 2024-11-26
remove tabBarIconStyle from material-top-tab-navigator options

imranbarbhuiya:
This option was removed in https://github.com/react-navigation/react-navigation/pull/11548


satya164 (approved):
Thanks!

PR #1391, opened 2024-12-30
add breaking-change headerTruncatedBackTitle to the doc

Freddy03h:
The change of `headerTruncatedBackTitle` to `headerBackTruncatedTitle` isn't mentioned in the upgrade helper and the v7 doc isn't updated.

https://github.com/react-navigation/react-navigation/pull/12090


satya164 (approved):
Thanks for the PR

PR #1394, opened 2025-01-02
Remove mention of jest-native in testing docs

jaworek:
Removed section mentioning `jest-native` package for setting up tests, as it is deprecated and is not used with newer versions of React Native Testing Library.


satya164:
Thanks for the PR. Closing since addressed by https://github.com/react-navigation/react-navigation.github.io/commit/7fc4ebc36e34ccf7210b40a421420d4bf82d2477

PR #1397, opened 2025-01-16
Fix ternary typo in `bottom-tab-navigator.md`

willemk:
Typo fixes for ternary operators I found while working with the example code. 

Also checked historical versions of the page but didn't find any problems there.


satya164 (approved):
Thanks for the PR

PR #1412, opened 2025-02-22
Update custom-navigators.md : Add missing variable

ahmed-khlifi:
In the `TabNavigator` example , `isFocused` Variable is using `index` variable to validate the condition. 
However, the `index` variable is not defined which cause an error of no existing variable.
old example :
```
state.routes.map((route) => ( // missing index
          <Pressable
            key={route.key}
            onPress={() => {
              const isFocused = state.index === index; // using undefined variable
...........
              }); 
           ..........
```


satya164 (approved):
Thanks for the PR!

PR #1414, opened 2025-03-08
Fix syntax error in tabBarPosition example

bkeepers:
The example uses a ternary operator and had a `?` where the `:` should be.


satya164 (approved):
Thank you!

PR #1415, opened 2025-03-08
Fix typo in auth-flow.md

mchisolm0:
# Change
I changed "...we're have..." which is the contraction form of "...we are have...". The proposed change is "...we have...", however, please feel free to suggest a different fix.

# Question
I read the text generated when I started the PR. I *think* I edited the file it was meaning (i.e. the versioned docs file). Please let me know if I misunderstood and I am happy to correct my error.


satya164 (approved):
Thank you!

PR #1418, opened 2025-03-26
Update docs about headerBackButtonDisplayMode and headerBackTitle

maciekstosio:
Changes to docs according to: https://github.com/react-navigation/react-navigation/pull/12500


satya164 (approved):
Thank you!

PR #1437, opened 2025-09-09
docs(iOS): Update docs for animationDuration in 6.x

kmichalikk:
Match the updated docs from types.tsx https://github.com/react-navigation/react-navigation/pull/12747, omitting "as of now".


kkafar (changes requested):
The same here. Let's revert the changes to not related files & we can proceed.

PR #1490, opened 2026-05-13
Add deferred deep linking section

BrtqKr:

### DIFF
--- a/versioned_docs/version-5.x/deep-linking.md
+++ b/versioned_docs/version-5.x/deep-linking.md
@@ -353,3 +353,15 @@ const linking = {
 ```
 
 Similar to the above example, you can integrate any API that provides a way to get the initial URL and to subscribe to new incoming URLs using the `getInitialURL` and `subscribe` options.
+
+## What if the app is not installed?
+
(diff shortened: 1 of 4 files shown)


satya164 (changes requested):
Thanks for the PR.

I'd remove changes from 5.x and 6.x since you're probably not testing with them, and we aren't actively maintaining these versions.

Before adding the links, I'd also request to change your examples to use `getInitialURL` + `subscribe` pattern instead of using the current pattern using `useEffect` and `ref` https://github.com/software-mansion-labs/react-native-detour/blob/30666c7c18fea8995550231b388a1aece275122b/examples/react-navigation/src/App.tsx#L62-L74, as this pattern doesn't integrate with React Navigation. So many of React Navigation's features are not supported (e.g., automatic redirection to the correct screen after auth, validation, automatic deep links and type inference in static config, etc.). It also requires a lot of boilerplate since the user needs to handle mappings themselves.

Here is an example of `getInitialURL` + `subscribe` https://reactnavigation.org/docs/deep-linking/#integrating-with-other-tools. Happy to answer any questions (you can ask @kkafar to add you to our discord if you want to chat directly).

BrtqKr:
@kkafar mentioned that last week, appreciate the feedback! We've already assigned one of our devs to work on that, so I didn't want to bother you 😅 – the [PR](https://github.com/software-mansion-labs/react-native-detour/pull/63) should be merged soon. Regarding the docs – removed v5 and v6.

satya164 (approved):
Thank you

PR #1504, opened 2026-07-24
fix: Change 'title' to 'label' in unstable header items example

christophby:
As [defined in the types](https://github.com/react-navigation/react-navigation/blob/e617cd23944247da3fa5e7b2827d662d6c8b9874/packages/native-stack/src/types.tsx#L852), ist must be `label` instead `title`


satya164 (approved):
Thank you

